PR #1, opened 2024-07-02
Add Warcraft - Orcs & Humans

tomojira:
Added Warcraft - Orcs & Humans (v1.22h, the final version). The file set matches with a clean installation from Redump:
`WarCraft - Orcs & Humans (USA) (Rerelease) (1996-10-13)`
And also matches with the file set in the GOG release.

DOSC contains sound config set to Sound Blaster 16.

Links:
- http://redump.org/disc/35137/
- https://www.gog.com/en/game/warcraft_orcs_and_humans
- https://www.pcgamingwiki.com/wiki/Warcraft:_Orcs_%26_Humans


PureDOS:
Just to make sure, Warcraft 1 does not use the audio tracks on the CD, right? A quick web search seems to suggest that (only MIDI in-game). Is the CD not at all required to play the game? No copy protection that checks the CD?

For consistency, an installation variant should also contain CD image(s) as a .CHD file(s), just like in regular game entries and not BIN/CUE files.

tomojira:
PCGamingWiki reports that the CD version had no DRM, and that does seem to be the case, I can play the game without any issue. I also notice now the wiki claims the game supports CD audio, but I don't see an option for this in-game, and yeah a search suggests that it isn't possible:
https://www.vogons.org/viewtopic.php?t=16554
One post in there theorized that the audio tracks on the CD were only for certain Mac installations.
But I'll change the installation variant to CHD, and double check everything in general.

tomojira:
I missed something on the PCGamingWiki page yet again, it does actually confirm that the CD audio was a Mac thing:
> Certain Mac OS (Classic) release supported CD quality streaming audio.
So it seems to fit as an Installation variant for DOS. I changed it to a CHD, and used the latest MakePureDOSDAT so it includes in_zeros and out_zeros. The current dates were also removed for the DOSZ.
One thing I'm not sure about is if the CHD should have a shorter name.

mariospicross:
Following the naming conventions seen in the rest of this DAT, it should just be `WarCraft - Orcs &amp; Humans.chd`

PureDOS:
The README doesn't really specify the naming of the disk images. Maybe keeping the naming from redump is a good thing, or maybe it is not. One thing though, "Warcraft" not "WarCraft" :-)

mariospicross:
That's true, and it could help users know exactly what CD they need. Worst case scenario we can just rename it later, checksum stays the same after all

PureDOS:
Thanks for the contribution!

I contemplated removing "PATH.WAR" because it references a D: drive that won't be available and a user who might install it from some other CD drive letter would end up with a different file but I left it in. The game seems to run fine without it existing but I can't say for sure.

PR #3, opened 2024-07-03
Add Caesar

mariospicross:
USA GOG files based on 1996 Sierra Originals re-release
.dosc sets sound config to Sound Blaster 16

[GOG](https://www.gog.com/en/game/caesar)
[Redump](http://redump.org/disc/103931/)


PureDOS:
Thanks for this! I changed the time stamps to make them match the files on the original 1995 CD, renamed INSTALL.EXE to SETUP.EXE (because the CD version uses that) and switched the BIN/CUE pair to a CHD with track listing.

PR #4, opened 2024-07-04
Add Capitalism Plus

mariospicross:
Installed from Redump European CD
.dosc sets sound to SoundBlaster 16 and MIDI to SoundBlaster Pro - OPL3
(The checksum actually stays the same on some of the files it changed the modified date of, so the comments I made about maybe having to create patches are null) 

[Redump](http://redump.org/disc/92025/)


PureDOS:
Thanks for the submission!

I modified the .DOSC file to move the 2 files into a SOUND subdirectory. I think the game is expecting that?

PR #5, opened 2024-07-04
Add Jazz Jackrabbit + Holiday Hare + HH95

mariospicross:
Main DAT and installation variant are from the [Enhanced CD-ROM](http://redump.org/disc/30629/) packaged with Jazz Jackrabbit 2
v1.3 variant is from [mail order CD-ROM](http://redump.org/disc/62977/)
Original website with shareware can be found [here](https://web.archive.org/web/20031207041306/http://jazzjackrabbit.com/download2.htm), links are dead though
Can be found [here](https://www.classicdosgames.com/game/Jazz_Jackrabbit.html) instead
GOG not supported. Patches files with lots of bugfixes and improvements, information can be found [here](https://www.gog.com/forum/jazz_jackrabbit/information_about_the_jazz_1_patch). Unsure if a dosc should be made with these patches? Especially because an error in the CD release caused one of the levels to be almost impossible to beat
.dosc sets language to English and sound to Sound Blaster 16. The two shareware expansions don't have a language select.


PureDOS:
I added a CONFIG.000 file to the 3 DOSC files to skip showing the settings screen at the start (settings are still available under an options menu in the game).

As an experiment. I added a DOS.YML file with the following content to the DOSC of the 2 shareware episodes:
```YML
run_path: C:\JAZZ.EXE
run_input: (WAIT:200)
```
What this does is it auto runs JAZZ.EXE and it skips showing the command line help screen shown for 200 milliseconds. It seems the shareware episodes are English only, so skipping past setup seems fine. The startup menu can still be reached by holding down Shift (on a keyboard) or L2 (on a controller), which will cancel any auto starting.

For another experiment I put the shareware episodes up on [a release page](../releases/tag/shareware). The release assets didn't want spaces in the filename so we ended up with a bit weird filenames, but it works.

PR #6, opened 2024-07-04
Add Rogue

10kmotorola:
Based on release version 1.49
Incompatible with Steam release; `ROGUE.EXE` hashes differently than any other version available and there's no `MKOPT.EXE`


PureDOS:
I don't know that much about Rogue, but is there a reason for why you chose 1.49? According to https://britzl.github.io/roguearchive/ there's many more versions. Though I don't really know which one would stand out to feature in the DAT.

Also I built the DOSZ according the DAT but I got this lovely death screen 😆 😆 😆
image

mariospicross:
Huh. Well, I picked 1.49 because that is the last official version (which you should probably edit your reply to remove that link, it contains copyrighted material). In that case, would it just be best to DAT the steam release even though it doesn't have `MKOPT.EXE`?

PureDOS:
Ah, thanks for the clarification. I was confused by the sentence "Rogue 3.6 ... was the first version of Rogue ever widely released."

Any idea what we can do for the copy protection? I'm surprised that I find so little about it. This [page on the Computer History Wiki](https://gunkies.org/wiki/PC_Rogue) explains the existence of the copy protection:
>The PC version of Rogue was copy protected, and if you pirated it, you would be killed, and your gravestone would say 'Killed by Copy Protection Mafia'.
> The copy protection was easily defeated by Copy II PC.

"Copy II PC" seems just floppy duplication? Not really sure how that applies to our version :-)

mariospicross:
From what I can gather, Rogue was initially developed on Unix, and the final version that was released on DOS was Unix 5.0.2, which coorelated to DOS 1.49. Since it was ported by a different company it seems they used a different version naming scheme. Anyway, I'll see what I can do about the copy protection :)

mariospicross:
I'm unable to find a way to remove the copy protection, including using Copy II PC. Would creating an patch file with the EXE Steam released be okay? I've tested it and it works.

PureDOS:
So Steam basically has a cracked version of 1.49? Then yeah, I think that's what we should do, original in DOSZ, patch in DOSC.
I guess we're not sure what the copy protection actually does? It's kinda funny, I would have guessed a game like Rogue would be documented more thoroughly.

10kmotorola:
I added empty .OPT and .SCR files. Apparently the game doesn't store your high score if there's no empty .SCR file. But from what I can tell copy protection just increases monster damage by a ton which obviously makes the game almost impossible

PR #8, opened 2024-07-04
Add Theme Hospital

mariospicross:
Main DAT from [Beta Five 13-5-9](http://redump.org/disc/15805/) Redump disc
Variant DAT is the result of extracting the CD image contents into the game directory, allowing the game to be launched without the CD image. Doing this makes it identical to the [GOG release](https://www.gog.com/game/theme_hospital), minus a few TXT files that have been excluded
Installer copies files without including dates, so those have been excluded
.dosc sets language to English, sound to Sound Blaster 16, and install type to max


PureDOS:
It seems the DOSC files are missing SOUND\DIG.INI and SOUND\MDI.INI?

Also, I'm wondering if we're getting anything out of using the MAX installation when the CD image is part of the DOSZ anyway. Generally an emulator will have an extremely fast CD-ROM drive so we shouldn't be slowed down by CD reading very much.
Though I can imagine a game behaving different depending if it thinks it reads from a CD or a hard disk.

Also we should keep HOSPITAL.CFG in the DOSZ because otherwise the game isn't playable. The idea is that games should always be usable without any DOSC, as long as the user is willing to do their own configuration and/or do whatever is required by copy protection schemes (i.e. enter a code from a code wheel).

mariospicross:
I'm unsure about the max install thing. It doesn't hurt, just wastes some extra space.. It might be worth just in case things behave differently. Other than that things should be fixed now.

PR #9, opened 2024-07-04
Add Warcraft II - Tides of Darkness + Beyond the Dark Portal

tomojira:
Added v1.33 of Warcraft II and its expansion, installed from Redump:

- [WarCraft II - Tides of Darkness (USA) (Rerelease) (1996-10-05)](http://redump.org/disc/42873/)
- [WarCraft II - Beyond the Dark Portal (Expansion Set) (USA) (Rerelease) (1996-10-09)](http://redump.org/disc/42787/)

The DOS version of Warcraft II hasn't been rereleased on modern platforms (the Battle.net Edition for Windows is on GOG). So this is all of the original files. All variants of the latest versions on Redump have the same data track as well.

The DOSC contains the CD path and sound config (Sound Blaster 16), as well as `INSTALL.HST`, which is just a history file showing the drive it was installed to. Including this may be too conservative, maybe it could just be deleted, I don't know if it would really be used for anything.


tomojira:
Sorry, I just realized I forgot to actually delete most of the DOSC files from the DOSZ. These are gone now:
```
DRIVERS/DIG.INI
DRIVERS/MDI.INI
INSTALL.HST
```

PureDOS:
I did some minor adjustments:
- Moved WAR2.INI into the main DOSZ. It makes the game playable without DOSC.
- Removed INSTALL.HST. The only place the character sequence "INSTALL.HST" exists is inside SETUP.EXE and I think it's for generating the file. So I'm quite sure the game does not need it. If it were needed, it would need to be in the DOSZ file.
- Renamed the CHD files in the ZIP to match re-dump (we started doing that for other games added lately)

I'm not quite sure about the last part, because the file names are so incredibly long :-) I also noticed that sometimes they are shown as shortened file names with just 8 characters, but that should be fixed in the emulator. But extremely long names might just not fit on the screen so it isn't great either...

PureDOS:
Changes were made in #18 and #34

PR #10, opened 2024-07-05
Add Ultimate Doom

tomojira:
Added Ultimate Doom v1.9, installed from Redump:

- [Ultimate Doom, The (USA)](http://redump.org/disc/4427/)

All of Redump's variants install the same files. Some also include the Windows 95 version installer, but this one is only DOS (plus a shareware installer for Heretic).

Many files were removed since they weren't included by GOG:
```
DMFAQ66A.TXT
DMFAQ66B.TXT
DMFAQ66C.TXT
DMFAQ66D.TXT
DWANGO.DOC
DWANGO.EXE
DWANGO.SRV
DWANGO.STR
HELPME.TXT
ORDER.FRM
README.TXT
```

`DM.EXE` was also removed because GOG uses a later version (v1.5) that came with Final Doom, Heretic, and Hexen. Maybe `DM.DOC` could also be removed then, since it's kinda pointless on its own, I was just conservative with it since it's the same with GOG.

The DOSC includes the defaults for controller type (keyboard only) and sound (Sound Blaster 16).

The problem here is the Steam version only has `DOOM.EXE` and `DOOM.WAD`, not even `SETUP.EXE`, so players wouldn't even be able to change their control scheme easily. For that reason I excluded Steam, and just went for parity with GOG. But I'm definitely not sure if this is the right move, and can change it to whatever is preferred.


mariospicross:
Probably worth changing the dosc to be mouse + keyboard since you can still play keyboard only like that and it's the way 99% of people will want to play

tomojira:
Sure, changed it to mouse + keyboard.

PureDOS:
Two things I want to talk about before merging this.

Controller changes:
DOSBox Pure comes with predefined gamepad mappings which map the default keyboard controls to an XBox360 style gamepad. Do you think this could this break this mapping?

Removal of DM.EXE
This is the startup EXE for multiplayer, right? DOSBox Pure along many other variants of DOSBox support tunneling of IPX and modem communication over the internet. Multiplayer is such a big part of Doom's legacy so I don't think we'd want to remove it just because the GOG release doesn't have it. We can't always match Steam/GOG. Or maybe I'm confused as to what DM.EXE does, if it's just some kind of readme-viewer then I don't think it's essential.

tomojira:
Mouse + keyboard setting doesn't seem to affect the gamepad controls. Moving the mouse around will affect your view and move you a bit, and clicking will shoot, but it doesn't remove keyboard movement and shooting. Actually, I see now that the default gamepad controls already have perfect twin stick control with RB to shoot, it feels pretty nice.

DM.EXE is indeed used for multiplayer:
https://doomwiki.org/wiki/DeathManager!
If it's important, I don't have a problem adding it back. Same with any other Doom/Heretic games where it got removed.

PR #11, opened 2024-07-06
Add Doom II + Master Levels

tomojira:
Added Doom II with the Master Levels for Doom II expansion, installed from Redump:

- [Doom II (USA) (v1.9)](http://redump.org/disc/62840/)
- [Master Levels for Doom II (USA)](http://redump.org/disc/21230/)

Maximum Doom also comes along with Master Levels, but I didn't include it because it seems to require Doom 1 and Heretic in addition to Doom 2. Not sure how things like that will be handled.

DOSC includes controller type (mouse + keyboard) and sound (Sound Blaster).

Unfortunately the Steam version is missing `SETUP.EXE`, similar to Doom 1. GOG is missing many files, but also `DOOMIT.BAT` and `DOOMIT.CFG` are very slightly different (CD install has an extra empty line at the end of both), and that's the launcher for Master Levels. So I just went full CD install with all files, but again not really sure if it's the best option.


tomojira:
I split Doom II into its own entry, and removed some files in the process so it's now compatible with GOG. There's also a DOS.YML to auto-run the game.

Master Levels is gone for now, but can be added back later if there's a system to handle Doom wads and such.

tomojira:
v1.9 doesn't seem to come with DeathManager. Not sure if DWANGO is usable at all or if it should be readded:
https://doomwiki.org/wiki/DWANGO

tomojira:
It's odd the Doom II CDs don't come with DeathManager, it seems like all Doom/Heretic games should have it, so I added a copy of v1.1. It's included with the Doom v1.9 shareware:
https://www.doomworld.com/idgames/idstuff/doom/doom19s

tomojira:
Master Levels seem to work well with the new parent system, so I've added them back.
One annoyance in this case is the paths in `DOOMIT.CFG` need to be edited to `C:\` and `C:\WADS\`. Or the default file could be kept if Doom II was moved into `C:\DOOM2\`, but that could cause annoyances for other mods. And GOG and Steam have slightly different default files anyway.
`DOOMIT.BAT` is also slightly different for GOG and Steam, and it really isn't worth adding variants. Best solution I could think of was to use the DOSC to distribute the exact versions of these files. Which is redundant, since they're already required in the DOSZ, but couldn't really think of anything else.

tomojira:
From what I understand, `SETUP.EXE` offered multiplayer from the start. DeathManager was a later, more convenient multiplayer launcher that was made after Doom II, and bundled with everything from Heretic onward, including Ultimate Doom. But unlike Ultimate Doom, the re-releases of Doom II I checked didn't have it. These pages claim it came with v1.7a though:

https://doomwiki.org/wiki/Versions_of_Doom_and_Doom_II#v1.7a
https://doomwiki.org/wiki/DeathManager!#v1.1

I just felt the need to include it for consistency with every other Doom/Heretic game, and I thought it could future-proof it if anyone wants to use mods with `DM.EXE`, if possible?
Also fine with testing anyway.

PR #12, opened 2024-07-06
Add Final Doom

tomojira:
Added Final Doom, installed from Redump:

- [Final Doom (USA)](http://redump.org/disc/21184/)

Similar to the other Dooms, Steam is missing `SETUP.EXE`. GOG actually modifies `DOOM2.EXE`, `PLUTONIA.WAD`, and `TNT.WAD` for whatever reason. So this is another one that sticks with all the files in the CD install.

DOSC includes controller type (mouse + keyboard) and sound (Sound Blaster).


tomojira:
I split up the 2 games, since they use separate installers into separate folders and don't interact with each other. Now they both have YML files to auto-run.

(This could still be disregarded if there's a system to handle Doom wads, I just thought it was a nicer experience than the previous commit)

tomojira:
Actually Final Doom probably shouldn't fall under a parent system. It comes with `DOOM2.EXE`, and it's possible someone might own it without having a copy of Doom II, so they wouldn't be able to build the DOSZ.
Also I changed to a bug-fixed version from id Anthology, which has a different `DOOM2.EXE`. GOG also uses this version. Steam doesn't, so I added a variant for that.

PR #14, opened 2024-07-06
Add Caesar II

10kmotorola:
Admittedly this may be slightly overkill, but Caesar II can be played without the CD-ROM inserted, but doing so allows for optional custscenes to be played. Since it's not required, I've made the main DAT without the CD-ROM, and then variants for with the CD-ROM and one for its videos extracted. I'm unsure if only one of the "CD-ROM" or "Extracted" variants should exist, so just let me know what you think. It's bloat-y for sure, but I guess the variants DAT is prone to that in general..

Installed from [Redump](http://redump.org/disc/29911/) original European CD-ROM
Removes C2.BAT for GOG parity but adds it in the dosc because why not
.dosc sets sound to Sound Blaster 16 and includes DOS.YML that autoruns `C:\C2.BAT`
Extracted CD-ROM contents copied without file dates, so those have been excluded


PureDOS:
> Caesar II can be played without the CD-ROM inserted, but doing so allows for optional custscenes to be played.

Am I reading that correct cutscenes play only without the CD-ROM inserted?

10kmotorola:
No, cutscenes do not play without the CD-ROM inserted

PureDOS:
Do you think it would be worth basing this off [Caesar II (Europe) (Rerelease) (1996-04-25)](http://redump.org/disc/36147/) which seems to contain Version 1.1A and also has a smaller CD image because it doesn't contain any demos or trailers? It is from a few months after the 1.0 release but I can't find any information on what has been fixed for 1.1 so it's hard to say if we truly need it or not.
If we'd forgo GOG compatibility in the main DAT entry, I think we should just go for the full install (around 10 MB of game  plus 16 MB of cutscene videos). And then maybe 2 variants (GOG, just the CHD)?

10kmotorola:
V1.1A was actually never ported to DOS, so I've kept it as v1.0A. Other than that I agree with your suggestions and have implemented them. Note the GOG variant has the `game.gog` we talked about, so I wasn't able to test the DOS.YML worked, but it should be fine

PureDOS:
We found a 1.2 version for DOS on a later CD release. Will merge this PR and then add another commit with the 1.2 update.

PR #16, opened 2024-07-06
Add Sid Meier's Civilization

10kmotorola:
Installed from the 2 v474.04 3.5" floppy disks, the last widely produced floppy release. Variant v474.05 DAT is the result of placing the .EXE that MPS Labs distributed to upgrade your copy. Since it required a manual patch, I'm keeping the more widely available floppy release as the main DAT.

DOSC contains some TXT / BAT files only found in this floppy release, making it possible to manually upgrade older versions of the game with EXEs MPS Labs provided and still be valid for this DAT. There is no sound configuration; it asks you manually each time you start the game. DOS.YML runs the keys `141`, ticking VGA, Sound Blaster, and mouse + keyboard. Also includes patch to disable the copy protection quiz you get asked every 50 turns. 

(I didn't mean to delete the branch for the other PR, I have no idea how that happened LOL)


PureDOS:
[This Civilization Fanatics Center forum post](https://forums.civfanatics.com/threads/patch-474-05.621927/#post-14860383) says
> v5: The most recent failed attempt to repair the Advisory Bug. Note: There is no other difference between v3 and v5.

Is it even worth having the variant?

10kmotorola:
Hah, that's funny as hell. I think it's still worth including it since it's a valid, official variant of the game regardless of if they failed to fix the bug.

PureDOS:
I think we can remove the BAT files in the DOSC.
I'm not quite sure about having INTRO.TXT/INTRO3.TXT and READ.ME in the DOSC. We don't really own these files I guess? It is neat that only these 3 files make things wider compatible...

10kmotorola:
It is probably safer to exclude them yeah.

PR #17, opened 2024-07-06
Add Heretic - Shadow of the Serpent Riders

tomojira:
Added Heretic v1.3, installed from Redump:

- [Heretic - Shadow of the Serpent Riders (USA) (Rev 1)](http://redump.org/disc/4450/)

GOG, Steam, and the Europe CD-ROM almost completely match this file set. However, a couple other CDs had different versions of these files, so they were removed:

```
DM.DOC
DM.EXE
README.TXT
```

`LICENSE.DOC` was also removed for being missing from Steam.
`DM.DAT` was removed for being pointless on its own.

DOSC contains controller type (mouse + keyboard) and sound (Sound Blaster), plus a YML for auto-run.


tomojira:
Readded the DM files and README.TXT, using the versions that matched between the later CD-ROMs, GOG, and Steam.

tomojira:
This is admittedly very thin for a Steam variant, it's just `LICENSE.DOC`, so definitely open to changing it back. Just trying to be consistent with this now. I also don't have Steam's default config, so the DOSC is just the same as the main entry.
At least GOG doesn't need a variant, it's 100% compatible.

PR #19, opened 2024-07-07
Add Hexen - Beyond Heretic

tomojira:
Added Hexen v1.1, installed from Redump:

- [Hexen - Beyond Heretic (USA) (Rerelease)](http://redump.org/disc/25734/)

GOG and Steam file sets are almost the same. Only `LICENSE.DOC` and `README.TXT` were removed for being missing or changed with Steam.

DOSC contains controller type (mouse and keyboard) and sound (Sound Blaster) along with DOS.YML for auto-run.

The game supports CD audio, but including the CHD would break compatibility with GOG and Steam. So looking at what orchid has done with Caesar II, I added a CD-ROM variant, though I'm not 100% sure about this. The DOSC is the same except with music set to CD of course.

The expansion, Deathkings of the Dark Citadel, doesn't patch the main game but does require a copy of it. So maybe that could wait until a parent rom system.


tomojira:
Swapped things around so the CD-ROM variant is now just the main entry, with GOG and Steam variants added instead. GOG has files called `game.ins` and `game.gog` which are basically CUE and BIN, and the music is extracted as OGG files. Steam doesn't offer any CD music.

PureDOS:
We removed the CD from the main DAT entry due to the CD audio soundtrack being just a recording of the MIDI soundtrack  and also a large number of tracks are missing. According to https://doomwiki.org/wiki/Hexen_music the game has 36 tracks but only 21 CD audio tracks.

PR #20, opened 2024-07-07
Add Sid Meier's Colonization

10kmotorola:
Installed from European re-release [Redump](http://redump.org/disc/49894/) CD-ROM. That CD-ROM is version 1.0 (it's not actually called that, I can't find any version history for this game other than 3.0, so it can be changed to "Original" or something to that degree if you'd like) but MPS Labs distributed a 3.0 upgrade, being the only patch they released for this game. Since that's the version available on GOG, I've made it the main DAT. From what I can find, it was distributed for both English and German, but I could not get the German upgrade to work. You know more than me about DOS, so if you'd like to try it, it was distributed as `COLGER.ZIP` with the sha1sum of `56b5a960bea20df6c242c4c24a5fa345f4bc2b1d` (unsure if I can send it here so I'll play it safe). It had a weird install process and even though it echo'd that it succeeded, none of the files changed. It seems to come zipped with all of the files it changes, but all of the `.SS` files are `.DSS` (presumably for Deutsch). I thought changing all of those from .DSS to .SS and just overwriting everything in the COLONIZE directory might've worked, but it wasn't able to launch. Now that I'm thinking about it maybe one of us can try removing all of the .SS within the COLONIZE directory and just replacing them with .DSS files without renaming them? Okay that was long, sorry :)

Some of the files loaded during installation changed modified dates, so dates for those have been excluded.

The CD-ROM actually also contains Spanish which I've included, but it's not mentioned anywhere on Redump. It doesn't seem worth it to request an account on their site just to fix that, but I dunno.

DOSC sets sound to Sound Blaster and contains a DOS.YML that autoruns `C:\COLONIZE.BAT`

(I accidentally forgot to checkout the correct branch before pushing, so that's why I removed Civilization in the second commit, sorry for the confusion)


PureDOS:
Colonization shows the version number on the main menu. 
On that CD there is 2.25 (English), 2.26 (French, German, Spanish).
COLGER.ZIP (which contains COLUP30G.ZIP, a file almost identical to colup30gd.zip from the (nowadays offline but archived) Microprose ftp server) updates it to 2.31 (even though it contains a file HOTLINE.TXT that says the update has 3.0). Though I just left the .DSS files as is next to the .SS files, maybe something isn't fully working.
Soo... I think everything you compiled is pretty good. I saw talk of a 4.0 patch but it looks user made and I couldn't even find it. The German patch I'm a bit conflicted. .DSS and .SS seem different formats (maybe one is compressed?) but I'm not sure how to deal with it. Keeping just the .DSS variants seems to result in a non-launching game, so I wonder if COLUP30G is meant for a different base version of the game (COLOUP.TXT describing how to update indicates it can be used for both floppy and cd version).

PR #22, opened 2024-07-08
Add Quake

tomojira:
Added Quake v1.8, installed from Redump:

- [Quake (USA) (Rerelease) (1998-09-21)](http://redump.org/disc/29723/)

All v1.8 CD-ROMs unavoidably installed the Windows source ports along with the DOS version (despite what the readme claims), but I think I've deleted everything related to WinQuake and GLQuake.

A few docs were removed for better compatibility with all v1.8 CDs:
```
HELP.TXT
LICENSE.TXT
README.TXT
QUAKE.CNT
QUAKE.HLP
```

Quake only has CD audio, so the CHD was included. As a result, this isn't compatible with GOG or Steam.

DOSC doesn't have any config from installation or outside of the game. Just DOS.YML for auto-run.


tomojira:
Restored all files, removed the config file from the DOSZ and added it to the DOSC, and added a GOG variant, with its `game.cue` and `game.gog` files. Actually not sure, should GOG's CD images be converted to CHD for consistency, or left as they are for convenience?
Also I realized Steam does not actually have the DOS version of Quake.

PureDOS:
It sounds reasonable to keep the GOG CD image files as is for a specific GOG variant.

PureDOS:
http://redump.org/disc/29723/ Seems to be the link to the matching redump info page.

If we're removing some unrelated files I think we can get rid of a few more.
- GENVXD.DLL/MGENVXD.VXD - used by WinQuake (I think)
- Q95.BAT/QLAUNCH.EXE  - launcher for multiplayer in Windows
- PDIPX.COM/QUAKEUDP.DLL - Windows multiplayer components
- VQUAKE.EXE - a special version of the game for a video card I have never heard about

Even with all these files remove, the game is still playable over IPX (which works online with RetroArch's NetPlay feature):
image

tomojira:
Oops, I had linked the Europe version instead. That version does match with everything anyway.

I'm fine with removing those files too. The reason I kept them was because I checked a Total DOS Collection entry for v1.01, and saw that it had those without the WinQuake/GLQuake stuff. Or with VQuake, I kept it because it technically runs in DOS, but yeah that video card has apparently never been emulated.

Part of me even wants to keep everything, for preservation's sake. I just figured it was too much clutter in this case with multiple unusable Windows versions. And the readme suggests the DOS installer should only install the DOS stuff, though that wasn't the case. But what do you think?

PureDOS:
I think because the CD has to be included anyway for the soundtrack, we're OK to clean up the installation to only have the required executables. Everything we're removing is "preserved" on the CD image that is right there anyway, right?

tomojira:
That is true. Yeah, I'm fine with removing them if you want to.

PureDOS:
Changes to Quake were made via #34

PR #27, opened 2024-07-09
Add Mario Teaches Typing

10kmotorola:
Matches [Mario Teaches Typing (USA)](http://redump.org/disc/61680/) from Redump
Original file dates didn't copy during installation
Requires CD-ROM to be inserted

DOSC sets sound to Sound Blaster and auto-runs `C:\MARIO.BAT` in addition to running `(WAIT:100)(RIGHT)(ENTER)` to skip the registration information it shows upon boot.


PureDOS:
It's neat to see the `(WAIT:100)(RIGHT)(ENTER)` working, but I ended up switching to launching `MARCD.EXE` directly, as done by the MARIO.BAT file. I still added a tiny `(WAIT:50)` so the first resolution reported by the core to the libretro frontend is the game's 320x240 and not the DOS console's 640x400.

PR #28, opened 2024-07-09
Add I Have No Mouth, and I Must Scream

televandalist:
- Version is the original via Redump: http://redump.org/disc/82875/
- DOSC configures Sound Blaster and auto-runs `C:\SCREAM\s.exe`


PureDOS:
This seems to be having a bunch of issues for me:
- DOS.YML refers to a `C:\SCREAM\s.exe` which does not exist (I tried fixing it to `C:\S.EXE`)
- Next it didn't have sound configured in the .DOSC. I had to run INSTALL.EXE to configure it which generated 2 files `DRIVERS\MDI.INI` and `DRIVERS\DIG.INI`)
- Then when launching it, it aborted with "Can't open music resource file."

televandalist:
> This seems to be having a bunch of issues for me:
> 
> * DOS.YML refers to a `C:\SCREAM\s.exe` which does not exist (I tried fixing it to `C:\S.EXE`)
> * Next it didn't have sound configured in the .DOSC. I had to run INSTALL.EXE to configure it which generated 2 files `DRIVERS\MDI.INI` and `DRIVERS\DIG.INI`)
> * Then when launching it, it aborted with "Can't open music resource file."

Made changes to the .DOSC that fixes all of these issues. Should be good to go now.

PR #29, opened 2024-07-09
Add Mario's Time Machine

10kmotorola:
Matches the four v1.0.1 3.5" floppy disks. From what I can tell this was never released as a CD-ROM for DOS. Installation didn't copy file dates.

DOSC sets sound to Sound Blaster and auto-runs `C:\MARIO.EXE`


PureDOS:
Should maybe the sound configuration be part of the DOSZ (with explanation of the chosen options during the install)?
Because as of now the DOSZ standalone won't have sound...

10kmotorola:
I was thinking that too since the installed files don't include a SETUP.EXE or equivalent. But also I don't wanna lock the DAT behind sound config when its not needed to play the game. So honestly I don't know

PureDOS:
We could keep the 4 floppies in the main entry so `A:\INSTALL.BAT` can be run with the "Product Re-Configuration" option to set new sound settings.

10kmotorola:
Yeah, sounds good to me

PR #30, opened 2024-07-09
Add Super Mario World DX

10kmotorola:
DAT'd exactly as [officially distrbuted](https://web.archive.org/web/20021210100530/http://www.bbsoftware.net:80/english.html)
Copyright date at the bottom of the readme states `2001,2002`, so I've put the release year as 2001.

DOSC auto-runs `C:\SMWDX.EXE`


PureDOS:
According to the readme the game was in development/private beta until version 0.3, and then 0.4 was the first public release.
The SMWDX040.ZIP has most files dated in the evening of 2002-03-23, except "SMWDXEditor" which is from 2002-05-04. So I think 2002 makes more sense for release year.

PR #31, opened 2024-07-10
Add Mario Brothers VGA

10kmotorola:
No installation; game was expected to be just ran off the floppy. I've seen the game referred to as `Mario Bros. VGA`, but at the top of `MARIO.DOC`, the title was shown as `Mario Brothers VGA`. This is the shareware/freeware demo; there doesn't seem to be a registered dump archived. The only difference is that the full version allowed multiplayer.

DOSC auto-runs `A:\MARIO.EXE`


PureDOS:
I think it would be better to have the DOSZ be the extracted files. The game saves the hi-score table which means with the DOSZ containing just the disk image, the save file would contain the entire disk image, too.

PR #33, opened 2024-07-10
Add Sonic PC

10kmotorola:
Another one from BB Software that was [officially distributed](https://web.archive.org/web/20021210100530/http://www.bbsoftware.net:80/english.html) freeware. This is the demo version and I highly doubt a full version exists anywhere online.

DOSC auto-runs `C:\SONICPC.EXE`


PureDOS:
Do you know if one can make the game output sound? It looks like there's 2 WAV files with music in there...

10kmotorola:
No, I don't believe you can

PR #34, opened 2024-07-10
Fix Quake comments, update Warcraft 2 DOS.YML

tomojira:
Fixed Quake's comments to have the right version number. It's v1.08, not v1.8:
https://quake.fandom.com/wiki/Quake_Version_History

Also added `run_mount` to Warcraft 2's DOS.YML so it loads Tides of Darkness instead of Beyond the Dark Portal, since people will probably play those missions first. (sorry, forgot to do this before)


tomojira:
Throwing in a few other minor changes - fine-tuning the `run_input` for Quake and Doom

PR #38, opened 2024-07-11
Add Hexen II & Portal of Praevus + Hammer of Thyrion

tomojira:
Added Hexen II and its expansion, with PAK files taken from Redump:

- [Hexen II (USA) (v1.11)](http://redump.org/disc/83185/)
- [Hexen II Mission Pack - Portal of Praevus (USA)](http://redump.org/disc/21268/)

And added into Hammer of Thyrion, which allows the games to run in DOS:
https://sourceforge.net/projects/uhexen2/

The games will crash with default settings, but run with memory size set to at least 32 MB according to the readme (I used 48 MB because I saw a comment by the developer recommending it). The EXE also needs a parameter to stream music from the mounted CD.

Portal of Praevus needs a `-portals` parameter as well, but I found that it didn't work in the DOS.YML, even though it did from the command line. I could be making another dumb mistake that I'm not seeing at the moment. But as a workaround, I included a BAT file in the DOSC which is used to auto-start. I also added one with the base game, could be convenient for re-launching the game from the start menu after quitting.


tomojira:
Following the Quake change, I increased the resolution to 640x400. I also increased mem_size to 64 MB - though I don't know if that's actually necessary, I kinda just did it for consistency with Quake II, in case there's any problems later in the game.

Also fixed `run_mount` for Portal of Praevus so it actually loads its own CD audio.

tomojira:
Added a DOS.YML with `mem_size: 65536` to the base game's DOSZ to ensure it's playable. The DOS.YML in the DOSC remains for auto start.

PureDOS:
Hammer of Thyrion supports OGG music which could be used instead of the very large CD images. But I don't know how we could instruct users to get to bit-exact OGG files :-(

tomojira:
That is a problem. GOG has Hexen II with extracted OGG files so maybe those could be used. They don't sell Portal of Praevus though.

PureDOS:
I'm not sure either, that's why I merged it as is :-)
Only way I can think of would be to document exactly what OGG encoder was used with which settings. If we had multiple cases like this we could write a simple OGG encoder that makes exact OGG files from CHD files, that wouldn't be too hard. But just for this one game.....

PR #39, opened 2024-07-11
Add Quake expansions

tomojira:
Added these Quake expansions, installed from Redump:

- [Quake Mission Pack No. 1 - Scourge of Armagon (USA)](http://redump.org/disc/20166/)
- [Quake Mission Pack No. 2 - Dissolution of Eternity (USA)](http://redump.org/disc/20167/)

And also [Dimension of the Past](https://www.quaddicted.com/reviews/dopa.html).

The BAT files for Scourge of Armagon must be edited to change the path to `C:\quake.exe`. This applies to the GOG variant as well.
For Dimension of the Past, `dopa.bat` must be edited from `winquake` to `quake`.


tomojira:
Updates:
- Converted audio tracks with CHDtoOGG
- Added links
- Changed title to "Quake Episode 5 - Dimension of the Past", since that's what's used in the readme and on Quaddicted

tomojira:
Even more updates:
- Moved DOPA to a Mods dat
- `dopa.bat` no longer needs to be edited. There's just a `PLAY.BAT` file in the DOSC instead, auto-started by the parent
- `run_mount` removed from mission packs since it's now unnecessary

PR #40, opened 2024-07-12
Add Quake II & Mission Packs + Q2DOS

tomojira:
Added Quake II and its mission packs, installed from these Quake II: Quad Damage releases:

- [Quake II (USA) (Rerelease) (1998-08-25)](http://redump.org/disc/43204/) (v3.14)
- [Quake II Mission Pack - The Reckoning (USA)](http://redump.org/disc/51202/) (patches to v3.15)
- [Quake II Mission Pack - Ground Zero (USA)](http://redump.org/disc/9891/) (patches to v3.19)
- [Quake II Netpack I - Extremities (USA, Europe)](http://redump.org/disc/9892/) (patches to v3.20 - I ignored the other stuff with it)

I used PowerISO to open them and ran the installers in Windows. I think the final results match with GOG as well.

Anyway taking the files from `baseq2`, `xatrix`, and `rogue`, you can add them to Q2DOS to play them in DOS:
http://dk.toastednet.org/Q2DOS/

The base game seems to work fine. Unfortunately I couldn't get the mission packs working consistently - they did at one point, but I don't even know why. Now The Reckoning shows an error, while Ground Zero shows some text you can Esc past to start a new game, but it will put you in the base game instead. If I put them all in a combined DOSZ, they do work when launching them from the command line:

`q2 +set game xatrix`

`q2 +set game rogue`

But the same thing doesn't work with parent/child. Which is very similar to what happened with Portal of Praevus:
https://github.com/PureDOS/DAT/pull/38
So it could be a bug, or I'm just seriously missing something, I'm stumped, but still thought I'd submit this in case anyone could figure it out.


tomojira:
I found part of the problem - something went wrong in the "sram" files at some point. I deleted those and started fresh, and The Reckoning no longer shows an error. Instead it starts with some text to Esc past, similar to Ground Zero.

To work around that, I changed their `run_path` to the BAT files I made, and now they launch normally. I don't know why these would behave differently:

`run_path: C:\Q2.EXE +SET GAME XATRIX`

`run_path: C:\Q2.BAT` with the same content

But they do. Maybe there is some kind of general bug here, it's definitely similar to what happened with Portal of Praevus.

Another problem is if you try to skip the Xatrix logo after starting a new game in The Reckoning, you get a black screen for a long time until an error appears. Could be a problem with Q2DOS, but this does not happen with a combined DOSZ, only parent/child.

tomojira:
Increased the resolution to 640x400. Also added `mem_size: 65536` to prevent any potential issues, since I saw mention of that somewhere - though it seemed fine at the beginning of the game.

tomojira:
Following with Arena and Daggerfall, I removed `mem_size` here since it doesn't seem to be needed, at least at the beginning. I also tried 3dfx Voodoo graphics, but it looked weird and buggy for me, and I don't know how to configure graphics properly, so I'll leave that to someone who knows what they're doing.

PureDOS:
I realized Q2DOS wrote a 24 MB large "CWSDPMI.SWP" file into my save game. I think this means the game is using the hard disk to swap out memory that doesn't fit into RAM which could probably be avoided by adding 24 MB more RAM to the default 16 MB. We probably should keep an eye for things like that... I wonder if the emulator could somehow warn if a swap file is being created so we have a better chance at spotting things like that.

tomojira:
Ohhh, well that explains a lot, it was strange that the game was somehow playable with 16 MB. Now I see that the readme does confirm this - I saw this already at some point, but forgot about it.
```
1 - What's required?
---------------------------------------------------------
* Pentium 1.  At least a 200MHz to play it at ~20-30fps.
* At least 48 MB RAM.  More for larger mods.
* LFN TSR.  Use the included DOSLFN.ZIP (can be loaded high with LH DOSLFN.COM)
```

PureDOS:
For us, in the emulator, performance wise we aren't affected much because the changes made to the C: drive is kept in memory of the host PC so access will still be very fast, much much faster of course than it would have been on a harddrive made 30 years ago. But it will unnecessarily bloat the save file and also likely introduce issues with save states as well as memory inspections (i.e. cheats) because something important might disappear from memory. I assume games usually clean up on exit so we can easily miss the usage of a swap file. An easy way would be to open the game's .save.zip file while the game is running. Or quit the emulator without letting the emulated game properly exit and then look at the save.

tomojira:
Yeah, now I'll know to watch out for that. In fact I just checked Battlespire, and yep it creates a 40 MB swap file, that explains what was going on there.

PureDOS:
Should we remove the "+ Q2DOS" part?
I guess if we do, we'd also remove the "+ Hammer of Thyrion" part in Hexen II #38.

tomojira:
Sure, I'm fine with that. I added it to make it clear that a fan project is involved, and it's not entirely the official product. But I don't mind either way.

PR #42, opened 2024-07-12
Add SimCity 2000 Special Edition

10kmotorola:
Matches [SimCity 2000 Special Edition (USA) (Rerelease) (1998-04-14)](http://redump.org/disc/8274/) from Redump. GOG had a different EXE that didn't match any other version, so that's its own variant. This is basically a CD-ROM collection of SimCity 2000 and all of its expansions, but no difference besides that. Regular SimCity 2000 can be DAT'd eventually I guess, but for now this is all I'm doing. Also worth noting there was a `SimCity 2000 CD Collection`, which is just this before they renamed it to `Special Edition` (I think when they ported it to Windows?).

DOSC sets sound to Sound Blaster 16 and auto runs `C:\SC2K.EXE`. Also worth noting for when we figure out MIDI stuff: this game has Roland MT-32 support


PureDOS:
There is a special feature on the CD called "WillTV". Not sure it's worth to include the entire CD within the DOSZ, but it might be possible to copy just the videos and modify `WILLTV` in SC2000.CFG. Also I think the DOSZ should contain SC2000.CFG, there doesn't seem to be a SETUP of some sort to recreate it after the installation...

10kmotorola:
Agreed, that should be good now. When merging main do you mind also removing `FILEMODS.DBP` from `SimCity (1989) (Maxis) (Ancient Cities + Future Cities + Terrain Editor).dosz`? Thanks :)

PureDOS:
I changed the auto start executable to what seems the only one that makes the "Watch WillTV" button in the game's main menu work.

PR #45, opened 2024-07-13
Add Chex Quest & Chex Quest 2

tomojira:
Added Chex Quest, installed from Redump:

- [Chex Quest (USA)](http://redump.org/disc/4490/)

The installer said it's for Windows 3.1, but surprisingly worked fine in Windows 10.
DOSC has Keyboard + Mouse controls, Sound Blaster config, and auto start.

Also added Chex Quest 2, which was a free download back in the day. I got it from this page:
http://www.doomlegends.com/chexquest/tucq.html

It has the first game as its parent. DOSC contains auto start.


PureDOS:
I think this should really just be CHEX.EXE, CHEX.WAD and SETUP.EXE (107056 bytes). Everything else is Windows exclusive.

PR #47, opened 2024-07-13
Add System Shock

tomojira:
Added System Shock Enhanced CD-ROM, installed from Redump:

- [System Shock (USA) (En,Fr,De)](http://redump.org/disc/31372/)

I didn't see an option for max install, so this has the CD image included. However GOG and Steam do have extracted installs, and they match each other after deleting 1 file from GOG (`CURRSAVE.DAT`).

`CYB.CFG` has English language and Sound Blaster 16 configured during installation. This file is needed in the DOSZ to run the game. So the DOSC just has auto start with the enhanced version.
Config can be changed with `C:\INSTALL.EXE`.


10kmotorola:
I think as long as the configuration file can be generated from within the DOSZ (such as running a SETUP.EXE), it should be excluded, even if the game can't run without it. I could be wrong though

tomojira:
Ohh, that's a good point, I can change that.

tomojira:
After testing it, I found that generating a new `CYB.CFG` after installation didn't work, because the CD path wasn't set. It would have to be edited manually. Probably best to keep it in the DOSZ unfortunately.

PR #50, opened 2024-07-13
Add SimFarm

10kmotorola:
Adds v1.01 of SimFarm. Variant DAT is v1.3 which installs the downloadable update. Note that the README indeed does say v1.3, while from what I can find the main version is v1.01, not v1.1. Name entered during setup was "Maxine", the default.

DOSC sets music to Sound Blaster v2.0 or later and SFX to Sound Blaster 16. Auto runs `C:\SIMFARM.EXE`


PureDOS:
Is there a reason why the newer version isn't in the main DAT?

10kmotorola:
I actually have no idea lol, should be good now

PR #52, opened 2024-07-13
Add Elder Scrolls - Arena

tomojira:
Added The Elder Scrolls - Arena, matching with Redump:

- [Elder Scrolls, The - Chapter One - Arena (USA)](http://redump.org/disc/2812/)

In `INSTALL.EXE`, "Install game" is blanked out, but you can install it simply by configuring sound and exiting. It only installs 3 files - `ARENA.BAT`, `SOUND.CFG`, and `NAMES.DAT`. Everything is run from the CD.

`ARENA.BAT` is modified by sound config (I chose Sound Blaster 16), so it's in the DOSC with `SOUND.CFG`. `NAMES.DAT` doesn't have anything to begin with, and it's generated again when you save your game, so I left it out. DOSC also has auto start.

The game is unplayably slow with default config, so I added `cpu_cycles: fixed 20000` in DOS.YML. This setting is recommended by [PCGamingWiki](https://www.pcgamingwiki.com/wiki/The_Elder_Scrolls:_Arena#Run_in_DOSBox) and also used by GOG. Steam goes a step further and uses 25000, but I stuck with the majority.
Other settings suggested by PCGamingWiki either aren't used by GOG/Steam or aren't supported by DOS.YML.

GOG and Steam have the game available for free, and distribute it as files extracted from the CD, with a workaround in the DOSBox CONF files. But this workaround doesn't work with a DOSZ, so unfortunately variants for Steam/GOG haven't been added.


tomojira:
Added `mem_size: 32768`, since it's used by GOG, Steam, and ArenaSetup.

tomojira:
After testing again, I found that the game is actually running fine without `cpu_cycles` set. No idea what made it so slow the first time.

PureDOS:
The way it is written in the YML file
```yml
cpu_cycles: fixed 20000
```
will be ignored because the word "fixed" is not supposed to be there. So it will fallback to the default.

The default in DOSBox is called "auto". Which means that games that run under the more modern [Protected CPU Mode](https://en.wikipedia.org/wiki/Protected_mode) will have the emulated CPU run as fast as possible. This includes games like DOOM and most games made after 1993 or so. Older games (and some games made after 1993) that run under the older [Real CPU Mode](https://en.wikipedia.org/wiki/Real_mode) will have the emulated CPU run at a fixed cycle rate. Pretty much all protected mode games are able to scale their game speed dynamically being unaffected by the CPU. Real mode games though show the described behavior, they often run too fast or too slow depending on the CPU speed.

Original DOSBox used to fall back to the very low default of 3000 cycles for all real mode programs. Emulators that support DOS.YML on the other hand by default try to detect the year of a games release. This can be done by detecting the running game via file signatures or by finding the release year mentioned in the loaded content. So if the loaded content file is "Elder Scrolls, The - Arena (1994) (Bethesda Softworks).dosz" it will realize that the game was released in 1994 and it will choose a CPU speed appropriate for that year. That means Arena will be tied to 55000 cycles which is similar to a 60 MHz Pentium processor available in 1994.

If the chosen CPU speed is not appropriate, it should be overridden with the DOS.YML option `cpu_year`. For DOSBox it will end up with the following cycle numbers:

`cpu_year` | cycles
------------|--------
1981          | 900
1982          | 1400
1983          | 1800
1984          | 2300
1985          | 2800
1986          | 3800
1987          | 4800
1988          | 6300
1989          | 7800
1990          | 14000
1991          | 23800
1992          | 27000
1993          | 44000
1994          | 55000
1995          | 66800
1996          | 93000
1997          | 125000
1998          | 200000
1999          | 350000

If you think it plays fine now, then we can remove `cpu_cycles` from the YML and just use the 1994 default.

tomojira:
Oh, that is good to know! The CPU year functionality sounds really convenient too. I'll have to revise some of my other submissions. I have found a few games that ran too fast - Mega Man, Mega Man 3, and Panzer General. Actually I wondered if Arena was also running too fast, you move much faster than in Daggerfall, but someone who knows the game well may need to judge this.

tomojira:
Testing again, `cpu_cycles` and `mem_size` are definitely not needed to run the game, at least at the beginning, so I removed them. I also tried `cpu_year: 1991`, which would be the middle ground between the `20000` used by GOG and `25000` used by Steam, but there was little (if any?) reduction in game speed, so I guess it's fine as it is. If the issues mentioned on PCGamingWiki become apparent later, could always be amended.

PR #56, opened 2024-07-14
Add X-COM: UFO Defense

televandalist:
Uses Steam version, which is already optimized for DOSBox

https://steamdb.info/depot/7761/


PureDOS:
Would it be possible to make this also work with a original media? Either a Floppy or CD re-release. 
Generally we don't really look for "optimized for DOSBox", rather original files. If A digital store release matches an installation from original media, that's just a bonus.
Some of the files included in the DOSZ (zmbv) contain DLLs for DOSBox which certainly doesn't need to be included.

According to https://www.ufopaedia.org/index.php/Steam_Versions Steam ships the 1.4 version of X-COM: UFO Defense but from a brief look it does not seem to match the [1.4 CD-ROM re-release from 1997](http://redump.org/disc/34427/). I think making the main DAT entry be based on a CD or Floppy release (1.4 sounds like latest) would be best, and if the Steam version doesn't match we can have that in a variant.

televandalist:
Not a problem. Was still getting a feel for things when I did this. I'll redo it with the CD-ROM version in the main one and Steam version in variants.

televandalist:
I'll do the same for Terror from the Deep

PR #58, opened 2024-07-14
Add Neuromancer

televandalist:
- 5.25" floppy dump that is simply an executable file and two DATs
- DOSC includes NEURO1.DAT, which was taken from the dump after mouse support and EGA were configured. NEURO1.DAT is also used for save games.
- Game auto-started on its own without having to configure anything else


PureDOS:
I think NEURO1.DAT needs to be in the DOSZ, too. Because games should always work without a DOSC.
Also it would probably be better to use a IPS patch because NEURO1.DAT seems quite large to just host on our own.

televandalist:
> I think NEURO1.DAT needs to be in the DOSZ, too. Because games should always work without a DOSC. Also it would probably be better to use a IPS patch because NEURO1.DAT seems quite large to just host on our own.

What would I be running NEURO1.DAT against to generate the IPS patch?

PureDOS:
If nothing needs to be different in the DOSC from the original NEURO1.DAT file copied during the installation, then nothing, no patch is needed.
But if there is some settings that we want to pre-set to make running the game easier with the DOSC which modifies NEURO1.DAT, it would be a IPS patch from the original installed file to the modified post-configuration file.

PR #60, opened 2024-07-14
Replace Grand Theft Auto

televandalist:
The biggest issue with the existing entry for GTA is the CHD. I haven't been able to reproduce with a matching MD5. The closest I found is Redump's dump of the Director's Cut version of the game; the file tree is nearly identical. This PR replaces existing info in the DAT with info generated from extracting data from this disc: http://redump.org/disc/30124/ + a CHD made from a bin/cue of the disc.

DOSC file was not touched.


PureDOS:
Was updated in 9ce885a

PR #61, opened 2024-07-14
Replace Cyberbykes

televandalist:
Could not reproduce the MD5 for the CHD, so I replaced info in existing DAT with info for the Redump dump of the game: http://redump.org/disc/2298/


PureDOS:
Was updated in b5e9487

PR #62, opened 2024-07-14
Add Elder Scrolls II - Daggerfall

tomojira:
Added Daggerfall, installed from Redump:

- [Elder Scrolls, The - Chapter II - Daggerfall (USA) (Rerelease) (1997-06-26)](http://redump.org/disc/68917/)

This release has v1.07.213, the final version, and also seems to include the CompUSA Special Edition content mentioned here:
https://www.pcgamingwiki.com/wiki/The_Elder_Scrolls_II:_Daggerfall#Patches

I encountered several different problems trying to get this to work. First of all, if you run the game from CD (necessary by default), it seems to be unplayable in DOSBox - it will always hang at a black screen after the opening movie. So it takes some more effort to set up a no-CD install. But as a bonus this greatly reduces the file size.

First I did a max ("huge") installation. This wasn't possible from a DOSZ in DOSBox Pure, because the combined size of the BIN/CUE and the install seemed to be too large for the virtual drive. So I used DOSBox-X to do it.

Secondly, all files from `ARENA2` on the CD need to be extracted into the installation's `ARENA2` folder. I used the `xcopy` command in DOSBox-X, or you could use PowerISO or something similar.
Specifically these 18 files are added:

```
ANIM0000.VID
ANIM0001.VID
ANIM0002.VID
ANIM0003.VID
ANIM0004.VID
ANIM0005.VID
ANIM0006.VID
ANIM0007.VID
ANIM0008.VID
ANIM0009.VID
ANIM0010.VID
ANIM0011.VID
ANIM0012.VID
ANIM0013.VID
ANIM0014.VID
ANIM0015.VID
DAG2.VID
PACKED.DAT
```

Then these two lines in `Z.CFG` must be edited like this:

```
path c:\arena2\
pathcd c:\arena2\
```

Now you can launch the game without a CD using the command `FALL.EXE Z.CFG`. However, I found that `run_path: C:\FALL.EXE Z.CFG` didn't work in DOS.YML - RALibretro would crash immediately. So I made `FALL.BAT` with the same command, which does work for some reason.
DOS.YML also has `cpu_cycles: fixed 50000` and `mem_size: 65536`, which are used by GOG, Steam, and DaggerfallSetup. PCGamingWiki recommends 32 MB mem_size instead but I went with 64.
The DOSC otherwise includes Sound Blaster 16 config.

Note that there are 2 files, `BIO.DAT` and `RUMOR.DAT`, that appear to be randomly generated when installed. But they're generated again when you start the game, so they're safe to leave out.

An easier alternative is to simply use GOG, which has the game for free, and already has all of that set up.
https://www.gog.com/en/game/the_elder_scrolls_chapter_ii_daggerfall
It doesn't completely match with the CD though. Notably it's missing one of the books, `BOK10000.TXT` - though there could be a reason for that, I have no idea. But I added GOG as a variant.
GOG's default sound config caused an error, so the DOSC has the same as the main entry.

Steam also has the game for free, but its `DAGGER.EXE` is different than GOG and the CD, and it appears to be missing some content in `ARENA2`. So I didn't bother with that.

As a final option, I added a variant for DaggerfallSetup, which has options to add unofficial fixes, quests, and utilities:
https://wiwiki.wiwiland.net/?title=Daggerfall_:_DaggerfallSetup_EN
I used the default setting, which adds all unofficial fixes but no quests or utilities. Like GOG, the default sound config caused an error, so I used the same as the main entry.
I'm actually not sure how community patches like this will be handled in the dat, so feel free to ignore this one if it isn't okay. But I could see people wanting this to be linked on RA for example.


tomojira:
Testing again, the game seems to run fine without `cpu_cycles` and `mem_size` set, at least at the very beginning.

But according to PCGamingWiki:
>Using "cycles=max" will cause heavy movement and turning issues. "cycles=fixed 50000" is recommended.

PureDOS:
What about these files on the CD? Are they not needed?
- /DAGGER/ARENA2/BIO.DAT
- /DAGGER/ARENA2/RUMOR.DAT

Would it maybe be safer to also include them in the DOSZ?

PureDOS:
The free GOG release is almost identical. It is missing these files:
- /ARENA2/BOOKS/BOK10000.TXT
- /ARENA2/RUMOR.DAT (not listed in the PR submitted DOSZ DAT entry)
- /DAG213A.EXE
- /DAG213B.EXE
- /DAGGER.PIF
- /FIXMAPS.EXE
- /FIXSAVE.EXE
- /FIXSAVE.TXT
- /INSTALL.EXE
- /INSTALL.PIF
- /SETUP.PIF

It does have /ARENA2/BIO.DAT, which was not in the submitted DAT entry.

I feel like we should just remove the files not in the GOG release (we can assume the GOG release is well tested so they aren't needed), add BIO.DAT, and then make the two identical?

What remains is that GOG's Z.CFG has the line "CHEATMODE 0" added at the end.

What if we added BIO.DAT and change the description to be

> Extracted from GOG release or installed v1.07.213 from 1997 CD re-release (any size install but all files from ARENA2 on CD must be copied to C:\ARENA2). The file Z.CFG must be edited to "type dfall_huge", "path c:\arena2\" and "pathcd c:\arena2\". On GOG release the line "CHEATMODE 0" needs to be removed.

Searching the web for "RUMOR.DAT" I found https://pt.uesp.net/wiki/Daggerfall:Corrupted_Savegames, which says:

>Workarounds for Corrupted Savegames:
> Delete all ATF and ATM files from the ARENA2 directory, as well as the BIO.DAT, MAPSAVE.SAV and the RUMOR.DAT files, also located in the ARENA2 directory

So perhaps none of these files should be in the DOSZ after all...

PureDOS:
Maybe we also include Z.CFG in the DOSC, it's small enough, just as a hint for someone who might get lost in editing it with line breaks or something else and can't get it to match the hash.

tomojira:
I explained `BIO.DAT` and `RUMOR.DAT` in the first post, but they appeared to be randomly generated, or at least different between releases. And they get generated again when you start the game, so it seemed okay to remove them.

I'm reluctant to remove `BOK10000.TXT` in case that's a readable book in the game or something. But it's also missing from DaggerfallSetup, intended to be a definitive community setup, so maybe it's okay. The other files probably don't matter, except for `FIXSAVE.EXE`, which GOG does have, but a later version of it. It is missing `FIXMAPS.EXE` so not sure about that one.
Since it's free, maybe GOG could just be used as the main entry, and forget about the CD.

tomojira:
Removed all the files mentioned, and therefore removed the GOG variant, I think this should be fine. `FIXSAVE.EXE` might be helpful sometimes, but everyone will still have it anyway.
Also added the required `Z.CFG` to the DOSC. `cpu_cycles` and `mem_size` aren't needed, at least at the beginning, so I removed those. And I added `run_input` for a skip, apparently I forgot to include that.

PR #63, opened 2024-07-14
Add Transport Tycoon Deluxe

10kmotorola:
v2.01.119 October build; from what I can find the latest build of the game. Some of the file dates were incorrect so those have been removed. Variant DAT adds [TTDPatch](https://www.ttdpatch.net/), a massive patch for TTD that pretty much everyone in the community used.

DOSC sets sound to Sound Blaster 16 and Sound Blaster AWE32


PureDOS:
Does "HS.DAT" stand for highscore? I don't think it needs to be included.

PR #64, opened 2024-07-15
Add Elder Scrolls Legend - Battlespire

tomojira:
Added Battlespire, installed from Redump:

- [Elder Scrolls Legend, An - Battlespire (Europe)](http://redump.org/disc/37479/)

According to Redump, the Europe release has the final v1.5, while the USA release has v1.2.

During installation, the video driver bypass was disabled (set to "normal"), video captions were enabled, and the childguard feature was disabled. These settings are stored in the DOSZ along with the CD path.

After installation, `SPIRE.CFG` must have the `path` edited to `C:\`.

Sound Blaster 16 config is in the DOSC along with auto start and `mem_size: 65536`, which I took from GOG. The game fails to detect the memory size correctly and warns you that don't have enough when you start it. This bug is acknowledged on this page:
https://www.dosbox.com/wiki/GAMES:The_Elder_Scrolls_Legend:_Battlespire
I used `run_input: (WAIT:1650)(ENTER)(WAIT:2100)` to skip past the warning.

GOG has a variant with its own CD image and extracted OGG music. DOSC has its default config. It doesn't have the erroneous memory warning, so it uses a different `run_input`.


tomojira:
I found more information on the memory error and how to work around it:
https://en.uesp.net/wiki/General:Playing_DOS_Installments_under_DOSBox#Battlespire
There's also info on using an `OBJMEMSIZE` option to avoid crashes, which is mentioned in `patch.txt` as well.
GOG has both of these changes. So I them copied over into `SPIRE.BAT` and `SPIRE.CFG` for the main entry, and put them in the DOSC. Not the DOSZ, so it can be built without making more changes. But the DOSC can override those 2 files and provide a bit better of an experience, hopefully. With this, the `run_input` is also the same as GOG.

tomojira:
The game actually seems to run fine without `mem_size` set, at least at the very beginning. It may be needed for performance later, according to some random comments I've seen online.

It does run extremely slowly when there's no sound configured.

tomojira:
As with other games, I changed my mind and removed mem_size until it can be verified that it's actually needed (in which case it would go in the DOSZ anyway, not the DOSC)

tomojira:
I've realized the game creates a 40 MB swap file, but this is part of SPIRE.BAT, so it still happens even when mem_size is increased. I guess it's okay as it is then.

tomojira:
Updates:
- Changed to WAITMODECHANGE
- Added links

PureDOS:
The DOSZ entry for SPIRE.CFG should probably be updated to match the file in the DOSC (with the added `OBJMEMSIZE	9000` line) and the <comment> updated to have that line added. Maybe we add the line at the very end of the file to make the explanation easier?

Now that DoDAT can use content embedded in the XML we could make our live easier for example if the file were
```
path         C:\
pathcd       D:\
bypass       0
captions     1
video        S3 Incorporated. Trio64
OBJMEMSIZE   9000
```
and listed (together with DOS.YML) in the DAT file like this:
```xml
		<rom name="DOS.YML" size="17" crc="9e2ade80" md5="5e4e9514ccbed110c66ed799f2b0c449" sha1="7ff552bf9a73a8ba6fe5089d140c63e3d6be659c" data="bWVtX3NpemU6IDY1NTM2DQo="/>
		<rom name="SPIRE.CFG" size="125" crc="31e0b5c3" md5="74b17c3becbe3987f8e556fd1a38ebf2" sha1="3212171238eeb0462f7b7e61f816c3869f340bbc" data="cGF0aCAgICAgICAgIEM6XA0KcGF0aGNkICAgICAgIEQ6XA0KYnlwYXNzICAgICAgIDANCmNhcHRpb25zICAgICAxDQp2aWRlbyAgICAgICAgUzMgSW5jb3Jwb3JhdGVkLiBUcmlvNjQNCk9CSk1FTVNJWkUgICA5MDAwDQo="/>
```

tomojira:
Done! That's a really nice feature.

A few other changes:
- Increased to high resolution and added this file in the DOSC
- Added `cpu_max_year: 2000` (maybe not necessary? I just add it to everything)
- Updated comments
- GOG DOSC also has its default config file

PureDOS:
One last thing, how are you feeling about having the `MAXMEM:70` in SPIRE.BAT only in the DOSC but `OBJMEMSIZE   9000` in SPIRE.CFG in the DOSZ? These seem somewhat connected so I feel like they both should be in the same place. And because both seem memory related, we probably would want them in the DOSZ... I think? Though that file feels almost too big to embed into the XML, it would be quite the blob
```xml
		<rom name="SPIRE.BAT" size="397" crc="8d871dcf" md5="81dd0d3611a95b4c06f1c86662999382" sha1="5f10adf9e8a2625a9ebe466fca85ef9526c99a21" data="QGVjaG8gb2ZmDQpDTFMNCmVjaG8uDQplY2hvIFRoaXMgdmVyc2lvbiBvZiBCYXR0bGVzcGlyZSBoYXMgYmVlbiBwYXRjaGVkIHRvIFYxLjUuDQplY2hvIEZvciBmdXJ0aGVyIGluZm9ybWF0aW9uLCBwbGVhc2UgZXhhbWluZSB0aGUgZmlsZSBQQVRDSC5UWFQgDQplY2hvIHdoaWNoIGlzIGxvY2F0ZWQgb24geW91ciBCYXR0bGVzcGlyZSBDRC4NCmVjaG8uDQplY2hvLiANCmVjaG8gU3RhcnRpbmcgQmF0dGxlc3BpcmUgVjEuNS4uLg0KZWNoby4NCmVjaG8gUGxlYXNlIFdhaXQuLi4NCmVjaG8uDQplY2hvLg0Kc2V0IGNhdXNld2F5PU1BWE1FTTo3MDtQUkU6NDA7TkFNRTpzcGlyZS5zd3ANCjpyZXBnYW1lDQpnYW1lIHNwaXJlLmNmZyAlMQ0KaWYgZXhpc3QgcmVwLmRhdCBnb3RvIDpyZXBnYW1lDQoNCg=="/>
```

tomojira:
Sure, I'm fine with that, it would make more sense - as long as that string isn't too big, but it seems okay to me.

PR #65, opened 2024-07-15
Add Elder Scrolls Adventures - Redguard

tomojira:
Added Redguard, installed from Redump:

- [Elder Scrolls Adventures, The - Redguard (USA) (Disc 1) (Install Disc)](http://redump.org/disc/2071/)

This disc is also needed after installation:

- [Elder Scrolls Adventures, The - Redguard (USA) (Disc 2) (Play Disc)](http://redump.org/disc/2072/)

The installer runs in Windows. Two versions can be installed, a software rendering version, or a Glide hardware-accelerated version. The latter looks better, but the former sometimes has better performance, according to this page:
https://en.uesp.net/wiki/Redguard:Glide_Differences

I couldn't figure out how to get the Glide version working, trying with the GOG release which uses it. It works with GOG's DOSBox setup, but I don't know how to replicate it. So for now I've just added the software version as the main entry.

The files are in `C:\Redguard` because the game gives a fatal disk error when it's at the root.

DOSC has Sound Blaster 16 config, auto start, and `mem_size: 65536`, which is used by GOG.


tomojira:
Added a DOS.YML with `mem_size` to the DOSZ, since the game refuses to run without it. The other DOS.YML still has auto start.

PureDOS:
As an alternative to running the installation on a Windows, all files can also be extracted by using a tool called [Unshield](https://github.com/twogood/unshield) which can extract the .CAB file extracted from the CD. There is a binary for Windows [here](https://github.com/ScoopInstaller/Binary/raw/master/unshield/unshield-1.5.1.7z) and it should be available in most [Linux distros](https://repology.org/project/unshield/versions). Extraction can be done with:
```sh
unshield -D 2 -d DATA1Extracted x DATA1.CAB
```

PureDOS:
We can either make a child "Elder Scrolls Adventures, The - Redguard (1998) (Bethesda Softworks) (3dfx Addon).dosz" which will be about 65 MB on its own or just add it to the main game and offer two DOSC files. Only issue might be that it requires [GLIDE2X.OVL](https://github.com/user-attachments/files/16474037/GLIDE2X.OVL.ZIP) to get included next to Rgfx.exe to get it to run. Do you prefer one of the two?

tomojira:
Hmm, I don't have much of a preference there, but I would probably lean towards combining them and using 2 DOSCs. Though one problem is GOG only has the hardware rendering files, and I believe Steam only has the software rendering, so they wouldn't be able to build it. I was thinking of having 2 separate variants to handle this, but that means many duplicated files and many lines, it's one of the longest line counts added so far.

televandalist:
I didn't notice this before, but the addition to `PureDOSDAT.xml` ends with `=======` instead of `</game>`.

tomojira:
Updates:
- Fixed the error mentioned by tele
- Added the 3dfx version combined with the software rendering version. Monster 3D v1.08 driver `glide2x.ovl` is needed: https://www.vogonsdrivers.com/getfile.php?fileid=158&menustate=2,1
- Converted audio tracks with CHDtoOGG
- Added GOG variant. It also needs the Monster 3D v1.08 driver, since the one provided by GOG doesn't seem to work outside of their custom DOSBox
- Added links
- Moved installation variant from the old file to the new one

PR #67, opened 2024-07-15
Add Crusader - No Regret

televandalist:
Installed v1.06 version using Redump-verified disc:
- [Crusader - No Regret (USA) (Rerelease)](http://redump.org/disc/82348/)

DOSC has keyboard + mouse controls, Sound Blaster config, subtitles enabled, large video option selected, and auto-start.


PureDOS:
Some issues:
- DOSZ doesn't run standalone (INSTALL.EXE can't be run to generate sound config or make the game run)
- SOUND.FLX can't be in the DOSC (file is large, so it likely contains copyrighted data and/or code)
- Seek cache (.SKC) files are emulator specific and don't need to be in the DOSC

I think the comment on the game just needs to explain all the options chosen at installation. Maybe if there are default options during install, the comment could just say that default options were chosen.

televandalist:
Fixed

PureDOS:
PR branch got broken and was replaced with #131

PR #68, opened 2024-07-15
Add Wasteland

televandalist:
- Installed using 2 x 5.25" floppies and selected EGA/RGB as the graphics/monitor combination.

- Disc info matches [ID 3108 (The Good Old Days)](https://www.goodolddays.net/diskimages/id%2C2233/)

- DOSC contains the GAME1 file from the installation, which is used for game saves. Game auto-starts as-is.


PureDOS:
I think GAME1 should stay in the DOSZ. It is too large and likely contains copyrighted data.

televandalist:
Fixed

PR #69, opened 2024-07-16
Add ...way too many Doom source ports

10kmotorola:
My sincerest apologies. I've tested all of these to make sure they work, which they do. A lot of source ports here didn't include CWSDPMI even though they required it, so I added the newest version `Version 7`. This is all* of the source ports/source modifications/executable hacks which support DOS that are on the Doom Wiki.

*I probably missed a few anyway


PureDOS:
What exactly are these? Mods or just source ports of original DOOM?

I wonder if all filenames listed in the DAT should be upper case, what do you think?

10kmotorola:
from DoomWiki.org:

`The Doom source code was released to the public in 1997. Although Doom was originally created for DOS, the release was of the Linux version, and the source code had to be ported back to DOS and to other operating systems. Hence the term "source port", which out of custom also is used for those modifications which are not technically ports to another platform. The term itself originates from the first source port, DOSDoom and its 0.2 version.`

`The original purpose of source ports was cross-platform compatibility, but shortly after the release of the source code, programmers were correcting old, unaddressed Doom bugs and deficiencies in their own source ports, and later began adding new features to alter gameplay.`

Basically, it's the basis a lot of custom WADs (mods) will need to run. tomojin and I's plan is to use the absolute minimum source port required for a WAD to run, which is the main reason I went so hard on the amount. For instance, if all a WAD needs is limit removal, we would use Doom128 since that does the bare minnimum we need. 

And I dont see a reason to change the casing; we haven't done so for any other post-DOS era DATs with modern casing

PR #78, opened 2024-07-19
Added XML data and DOSC for Alone in the Dark.

bcrew1375:


televandalist:
FYI: The year and comment lines start with four spaces and a tab instead of just two tabs

bcrew1375:
Fixed. Thanks.

PR #83, opened 2024-07-20
Add Ultima Underworld I & II

tomojira:
Added Ultima I & II, installed from this Redump-verified release:

- [Ultima Underworld I & II (USA)](http://redump.org/disc/47050/)

The CD install keeps a lot of files on the disc, including the executables, but I thought it would be odd to have a CD image with 2 games on it, so I extracted all the game files for each after installing them. I also considered using a floppy release instead, but they need patches for the final version.

DOSC has Sound Blaster Pro config and auto start. For some reason Ultima Underworld II's sound config from installation didn't seem to work, there was no sound, but I just replaced it with the same config from the first game.

I haven't added GOG variants because they have a single `game.gog` image that won't auto-mount. The default config also had no sound, though I could just copy over the working one again.


PureDOS:
DOSZ doesn't seem to run standalone without UW.CFG. The UWSOUND.EXE program is supposed to configure sound but it doesn't run without UW.CFG existing. So maybe we need to include the file in the DOSZ and add more install instructions how to get to that. We could just say to install with "No Sound".

tomojira:
Oops, sorry about that. That seems to be the simplest way to deal with it, I set "No Sound" during installation for both games, and added that UW.CFG file to the DOSZ.

televandalist:
> I haven't added GOG variants because they have a single `game.gog` image that won't auto-mount. The default config also had no sound, though I could just copy over the working one again.

I run into this on occasion. I just rename it to `game.iso` or whatever. There's typically another file that acts as a `.cue` file so I rename that, too. Then I change the filename inside the `.cue`.

tomojira:
I forgot about that - yeah, it should be easy to add the GOG variants then.

PR #84, opened 2024-07-20
Add Alone in the Dark 2.

bcrew1375:
Adds Alone in the Dark 2 matching Redump: http://redump.org/disc/66099/


televandalist:
Comment line doesn't have `</comment>` at the end

bcrew1375:
Fixed.

PR #85, opened 2024-07-21
Add Descent & level packs

tomojira:
Added Descent v1.4a, installed from the Descent Anniversary Edition CD-ROM:

- [Descent - Anniversary Edition (USA)](http://redump.org/disc/23720/)

There is a later version, v1.5, but it mostly just adds VR features. The CD releases that come with v1.5 are bugged and don't have music in the later levels:
https://cloakedthargoid.wordpress.com/pc-games-collections/descent-definitive/
So v1.4a seemed good enough to go with.

There are 2 official level packs, Levels of the World, and the Anniversary Edition levels. The latter gets installed with the base game, while the Levels of the World need to be extracted from the disc. They could both actually be combined with the base game, since they use the main EXE to launch, but I thought it might be better to keep it focused on the original game, with these level packs as child DOSZs. Especially Levels of the World, which has at least 100 of them.

GOG and Steam don't come with the level packs, and they match each other.


PureDOS:
Can we remove "SPANIARD.PLR"? Seems weird to have a player profile already existing.

tomojira:
Sure. It was added by the installer, but I'm not sure why, it was odd.

PR #86, opened 2024-07-21
Add Bedlam

televandalist:
- Installed using [Bedlam (USA)](http://redump.org/disc/37471/)
- Default settings used
- DOSC auto-starts the game
- Added the CD itself to the Variants DAT


PureDOS:
This game uses CD audio so I think the CHD should be included in the main DOSZ.

PR #87, opened 2024-07-21
Add 2400 A.D.

televandalist:
- Extracted from 5.25" floppy disk image that matches [ID: 756 from The Good Old Days](https://www.goodolddays.net/diskimages/id%2C756/)
- DOSC auto-starts and contains an .IPS for the executable that bypasses copy protection


PureDOS:
Does the .IPS patch remove copy protection?

televandalist:
> Does the .IPS patch remove copy protection?

Yep. Had to go back and double check.

PR #91, opened 2024-07-21
Add Rise of the Triad - Dark War

televandalist:
Site License CD-ROM - v1.3
- Installed from Redump's [Rise of the Triad - Dark War (USA) (Site License) (v1.3)](http://redump.org/disc/33505/)
- DOSC auto-starts
- Added disc to variants

Sound configuration used:
- Card type: Sound Blaster or compatible
- Port: 220
- IRQ: 7
- DMA: 1
- 16-Bit DMA: 5
- 4 Voices
- Stereo Sound FX
- 16-Bit Mixing


PureDOS:
Is there much desire to include these few custom levels? Loading these levels is quite cumbersome (need to go through setup, then the regular game can't be played until the setup is done again). I guess we could provide launch configurations to launch each of these level files, but unless they are considered essential/important (notorious? famous?) addons for ROTT it might be better left for perhaps a separate addon DOSZ?

PureDOS:
RAND0001.RTL is not included on the CD, it is randomly generated (differently) when running RANDROTT.EXE so I have excluded it from the DOSZ. With that, I have now also excluded the `RAND0001` launch config. Alternative would be to add launch configs to generate a new random level and another one to play it, but RANDROTT.EXE has a lot of options so it's probably best left for users to use it from the command line.
The ROTTSITE.RTC seems to be the default level pack for Comm-Bat when just running regular ROTT.EXE so I have also removed the launch config for that. ROTTCD.RTC seems to contain the same levels as ROTTSITE.RTC, but ROTTSITE has 10 or so more levels, so I removed the launch config for that as well.

PR #93, opened 2024-07-21
Add Descent II & Vertigo Series

tomojira:
Added Descent II - first installed v1.0 from this Redump-verified release:

- [Descent II (USA)](http://redump.org/disc/1947/)

Then patched to v1.2 with The Vertigo Series:

- [Descent II - The Vertigo Series (USA)](http://redump.org/disc/37289/)

I started with v1.0 instead of v1.1 because those later CDs introduce a remixed soundtrack - but also mess up the way it's played in-game:
https://cloakedthargoid.wordpress.com/pc-games-collections/descent-definitive/
So I wanted to include the v1.0 CD image for the original soundtrack.
Also, even though The Vertigo Series CD has audio tracks on it, they aren't actually usable, because the game outright requires the original CD to be inserted.

I went with a child DOSZ for The Vertigo Series levels, though they launch from the main EXE along with the main levels, so they could potentially be combined. I just figured people might want the original game on its own.

DOSC has CD audio configured and auto start.

GOG variant isn't included because it uses a single `game.gog` CD image that won't auto-mount.


tomojira:
The more I thought about it, the more I felt it would be more consistent to merge the Vertigo Series with the base game instead of making it a child DOSZ. It's not really the same as the first Descent, where the level packs feel like too much clutter, especially the 100 fanmade ones. This is more like a new official campaign, and it installed over the base game. So I've changed the dat to have it as one entry, similar to other expansion entries. Can be changed back of course if that's a mistake.

tomojira:
Updates:
- Added GOG variant. I don't know how I had a problem with it before, but it works fine now
- Added `cpu_max_year: 1998` to reduce the bobbing
- Changed to WAITMODECHANGE
- Added links
- Deleted `EREGCARD.INI`
- I wanted to increase the resolution, but it seems to be saved under the player's profile name, so can't really be included

tomojira:
More updates:
- Split Descent II and The Vertigo Series into parent/child again
- v1.2 patch for the base game: https://soggi.org/files/misc/game-patches/Descent-II_1.2_DOS.7z
- 3dfx package added (not made by Parallax Software, but approved by them): https://soggi.org/files/misc/game-patches/Descent-II_3dfxR6.7z
	- Needs `glide2x.ovl` from Monster 3D v1.08 drivers: https://www.vogonsdrivers.com/getfile.php?fileid=158&amp;menustate=2,1
	- Also needs DOS.YML with `mem_size: 65536`
	- 3dfx option added to DOSC
- Vertigo Series doesn't need its own DOSC

tomojira:
Even more updates:
- Changed to small size installation since the disc is included anyway
- Removed DESCENT.CFG from DOSZ (accidentally left it in last time)
- Changed patch links to the pages instead of directly to the files

tomojira:
I realized DESCENT.CFG is needed in the DOSZ or SETUP.EXE won't work properly (it will try to reinstall the game).
Complete instructions from the dat comment:

```
Installed v1.0 from CD.
Options:
	Small installation size
	High resolution movies
	Sound: Sound Blaster 16, Port 220, IRQ 7, 8-bit DMA 1, 16-bit DMA 5
	Music: SB Pro/16, Port 388
Then patched to v1.2, and 3dfx package added.
Needs glide2x.ovl from Monster 3D v1.08 drivers.
Needs DOS.YML with mem_size: 65536.
CD image converted with CHDtoOGG.
DESCENT.CFG is needed in the DOSZ for SETUP.EXE to work without reinstalling
```

Also added Setup Program option to the DOSC.

televandalist:
> GOG variant isn't included because it uses a single `game.gog` CD image that won't auto-mount.

;)

PureDOS:
I switched to the medium installation after all because I noticed the game copies 5MB of data (into the save file) on every level entry. Not quite an SSD killer but still not worth the 24 MB of larger installation (which compresses down quite well to only 5 MB in the DOSZ).

I also added a lower CPU limit specifically to the 3dfx version which otherwise has extremely fast up-down ship bobbing. Which made me wonder... It might be worth looking at https://cloakedthargoid.wordpress.com/descent-ship-bobbing/ it seems just 4 bytes of patch in the EXE, could be even a launch configuration in just the DOSC.

tomojira:
Oops... Sorry about that, something else to watch out for when checking install sizes.
Removing the ship bobbing would be great, but would it be okay to have such unofficial patches in a DOSC, would it set precedent for other games? I'm fine with it anyway.

PR #95, opened 2024-07-22
Add Star Wars - Dark Forces

tomojira:
Added Star Wars: Dark Forces v1.0 Build 1, installed from this Redump-verified release:

- [Star Wars - Dark Forces (USA, Europe)](http://redump.org/disc/9409/)

The largest install with all modules was used.

Files are in `C:\DARK\` because I found the sound didn't work when they were in `C:\`. Same with the GOG variant, except it gave an error and refused to launch.

The CD image needs to be included for the game to run - though there is a no-CD method, I found info that it breaks `SETUP.EXE`. GOG does use the no-CD method regardless.

DOSC has Sound Blaster 16 config along with some default config and auto start.


PureDOS:
Is there a reason to use the largest install as opposed to just use the smallest install? In emulators CD access shouldn't be measurably slower so I don't think we gain anything from a non-minimal installation.

tomojira:
Only reason is having more files exposed, and maybe being able to mod them more easily. But that's just theoretical.
I've been doing maximum installations for everything as a standard, but we can change to minimal if we want to be less redundant instead.

tomojira:
Updates:
- Changed to custom minimal installation size
- Changed to WAITMODECHANGE
- Added `cpu_max_year: 1997`
- Added links
- Added MT-32, General MIDI, and setup options
- GOG config is now the same as the main entry
- Moved GOG variant to Digital dat

PR #97, opened 2024-07-23
Add Ultima 0 - VII

televandalist:
- All seven games were extracted from a Redump-verified disc of [Ultima Collection (USA) (Rerelease)](http://redump.org/disc/2179/)
- DOSC files auto-start their respective games
- I-V do not have music as-is. There are unofficial patches that add music which will be added to variants later.

Full list of games added:
- Ultima 0 - Akalabeth (1980) (Origin Systems)
- Ultima I - The First Age of Darkness (1987) (Origin Systems) [Date is not a typo](https://www.pcgamingwiki.com/wiki/Ultima_I:_The_First_Age_of_Darkness)
- Ultima II - The Revenge of the Enchantress (1982) (Origin Systems)
- Ultima III - Exodus (1983) (Origin Systems)
- Ultima IV - Quest of the Avatar (1985) (Origin Systems)
- Ultima V - Warriors of Destiny (1988) (Origin Systems)
- Ultima VI - The False Prophet (1990) (Origin Systems)
- Ultima VII - Part Two - Serpent Isle (1993) (Origin Systems)


tomojira:
Testing these next. So far I've only done 2 games but here's what I've found.

### Akalabeth

`SOUND.CFG` can be removed from the DOSZ and added to the DOSC - it doesn't match with the one on the disc. It can also be created again safely at any time with `SETUP.EXE`

The game crashes after choosing a lucky number. I found an explanation for this here:
https://www.dosbox.com/comp_list.php?showID=717&letter=A
To fix it, you have to get a later version of CWSDPMI from here:
https://sandmann.dotster.com/cwsdpmi/
Then create a BAT file to run the game like this:
```
cwsdpmi -x
ak
```

A CPU setting can be added for consistency. I tried `cpu_year: 1983` and it seemed fine (same as what GOG uses, 1500 cycles)

I tried a General MIDI option but only Sound Blaster seemed to work.

#### GOG Variant

I found that this doesn't match with the free download from GOG. That version looks similar to the main entry, except the executables are different.

The current GOG variant does match with this TDC entry: `Akalabeth- World of Doom (1996)(California Pacific Computer Co., Finire Dragon) [Adventure, Role-Playing (RPG)]`

### Ultima I

The speed could be reduced, it's not very noticeable but `cpu_year: 1981` or 1982 seem a bit better for the sound effects and battle animations. 1981 moreso but then it might feel a bit more sluggish and unresponsive. (GOG uses 500 cycles)

At the default speed, the WAITMODECHANGE isn't needed, the game already starts at 320x200, so it skips past the intro. But after reducing speed, it does work out as it is.

#### GOG Variant

Almost matches with the main entry. `MOSLO.COM` and `ULTIMA1.TXT` could be removed, so there wouldn't have to be a variant.

televandalist:
Updated Ultima 0
- Replaced `CWSDPMI.EXE` with the newer version [here](https://sandmann.dotster.com/cwsdpmi/)
- Removed `SOUND.CFG` from the DOSZ and added to the DOSC
- Added `cpu_year: 1983` to `DOS.YML`
- Created `AK.BAT` which contains the following:
```
cwsdpmi -x
ak
```
- `AK.BAT` is now used to launch the game

Ultima 0 - Akalabeth (GOG)
- I don't know what the hell happened here. I redownloaded the GOG version and redid the whole thing from scratch.

tomojira:
Tested Ultima II to VI, these all built successfully with DoDAT and worked fine for the most part. I just have notes on adding CPU years, tweaking the run_inputs, and potentially merging some of the GOG variants.

### Ultima II

- I think `cpu_year: 1981` might be best to slow down the game here, at 1982 and higher you can see the water animation is too fast. But as always the tradeoff is the game feels less responsive
- WAITMODECHANGE could be removed since it cuts off the intro
- Unlike the other games, GOG isn't close to matching the main entry, there are many differences. `MOSLO.COM` could still be removed since it's a unnecessary program from the 90s that wouldn't have existed with original releases of the game, but doesn't matter too much either way

### Ultima III

- `cpu_year: 1981` seems like a good speed again
- WAITMODECHANGE could again be removed since it cuts off the intro
- Similar to Ultima I, `MOSLO.COM` and `ULTIMA3.TXT` could be removed and GOG would match without needing a variant

### Ultima IV

- `cpu_year: 1981` seems best for the animations... but can definitely feel sluggish sometimes
- WAITMODECHANGE is fine here. If 1981 is used, (WAIT:8000) could be added on top of it to get through that slow start
- With only `MOSLO.COM` removed, GOG would match the main version without needing a variant

### Ultima V

- `cpu_year: 1985` slows down the logo at the beginning. Nothing too noticeable
- No problem with WAITMODECHANGE
- GOG could match the main entry with only `MOSLO.COM` removed

### Ultima VI

- `cpu_year: 1985` is good for the TV speed in the opening. A bit slower than GOG's 3000 cycles
- No problem with WAITMODECHANGE. If 1985 is used, WAIT:4500 could be added to get the game started faster
- The `SAVEGAME\OBJLIST` file gets modified after playing, but it is needed to run the game, so maybe the default one could be included in the DOSC
- Unfortunately `OBJLIST` also prevents GOG from matching the main entry (along with `MOSLO.COM` again)

tomojira:
### Ultima VII: Serpent Isle

- `cpu_year: 1993` seems fine here. It looks a little fast but reducing as low as 1990 doesn't change it, and GOG doesn't use any lower than that (14000 cycles)
- WAITMODECHANGE can be removed, it skips a chunk of the intro
- MT-32: According to Vogons Wiki, the game has "Distorted/incorrect sounds on MT-32 'old'". So the key should be changed to `sound_mt32: CM-32L ROM v1.02 (1989) (Roland).dosz`
- Similar to The Black Gate, a DOS.YML should be included in the DOSZ with this, since the game can't be played without it (well, the UMB one doesn't seem to be strictly needed, but it's still recommended online):
```
mem_ems: false
mem_umb: false
```
- The new data parameter that lets it be generated by DoDAT could be included like this:
`<rom name="DOS.YML" size="32" crc="709d92a6" md5="4e1473d3eeae7d4a506e33972c967fae" sha1="c9830231e1b10fc93a274837ff7fc531ae19a0a6" data="bWVtX2VtczogZmFsc2UNCm1lbV91bWI6IGZhbHNlDQo="/>`
- Title could be changed to "Ultima VII - Part Two - Serpent Isle + The Silver Seed" to match with "Ultima VII - The Black Gate + The Forge of Virtue"
- GOG could match the main entry if `MOSLO.COM` and `SERPENT.TXT` were removed

tomojira:
Oh, `SAVEGAME\OBJLIST`  is still needed in the DOSZ for Ultima VI, since the game isn't playable without it. The problem is GOG, but this new embedded patch feature should allow GOG to match:

```
<rom name="SAVEGAME/OBJLIST" size="7539" crc="1f2bf9f5" md5="0dbebf2e8f97921cb89dd0c46c575c4a" sha1="910829de03001a29fb26988e5a501d8d10147d26" date="1991-09-12 07:03:00">
    <patch data="UEFUQ0gAG/QAAAB5AEVPRg==" size="7539" crc="106ff995" md5="f1c1f8e4c0535ef03581baa3f9e6d020" sha1="4fb89dd5a086e6247923bf80af9328cbc5c30a75"/>
</rom>
```

A few other things:
- I missed this before, but the GOG variant for Akalabeth could actually use WAITMODECHANGE to skip a short 640x400 part, it's a bit different than the main version
- The GOG variant for Ultima II is still missing `cpu_year: 1981`, and could have WAITMODECHANGE removed
- Ultima III still has `ULTIMA3.TXT`
- Serpent Isle DOSC hasn't been renamed to match the DOSZ
- Serpent Isle's DOS.YML from the DOSZ could also be uploaded to the yml folder in the repo, like these: https://github.com/PureDOS/DAT/tree/main/yml (just as another backup, though maybe it isn't as necessary anymore with the embedded data)
- Redump media links could be added

televandalist:
> Oh, SAVEGAME\OBJLIST is still needed in the DOSZ for Ultima VI, since the game isn't playable without it. The problem is GOG, but this new embedded patch feature should allow GOG to match:

Added it back in. Not sure what this "new embedded patch feature" stuff is all about, though...

> I missed this before, but the GOG variant for Akalabeth could actually use WAITMODECHANGE to skip a short 640x400 part, it's a bit different than the main version

It does have `WAITMODECHANGE` 🤔 

> The GOG variant for Ultima II is still missing cpu_year: 1981, and could have WAITMODECHANGE removed

Done. I must have edited `DOS.YML` directly from the DOSC instead of replacing it. Sometimes it doesn't actually save when that happens.

> Ultima III still has ULTIMA3.TXT

Removed. 

> Serpent Isle DOSC hasn't been renamed to match the DOSZ

Done.

> Serpent Isle's DOS.YML from the DOSZ could also be uploaded to the yml folder in the repo, like these: 

https://github.com/PureDOS/DAT/tree/main/yml (just as another backup, though maybe it isn't as necessary anymore with the embedded data)

> Redump media links could be added

I realized that I never actually added the installation variant. It was in my ROMVault DAT that I use for all this, though. Weird. Anyway, I added it.

tomojira:
Oops, sorry, I think I forgot to actually update that Akalabeth DOSC.

The "embedded patch" is a DoDAT thing in the latest version. It will try to identify the file that matches the info in the `<patch/>` tag, then apply a small IPS/BPS patch which is embedded in `data=` in Base64 format, and this patch will cause it to match with the info in the `<rom>` tag. It just needs to be formatted like this (closing slash removed from the end of the first `<rom>` tag, then `</rom>` added after the `<patch/>` info):

```
<rom>
    <patch/>
</rom>
```

PR #98, opened 2024-07-23
Add Star Wars - X-Wing + Imperial Pursuit + B-Wing

tomojira:
Added X-Wing with its expansions, installed from 3.5" floppy disks matching with these TDC-verified releases:

- Star Wars- X-Wing (1993)(LucasArts Entertainment Company LLC) [IMA][1440k][!]
- Star Wars- X-Wing - Imperial Pursuit v1.1 (1993)(LucasArts Entertainment Company LLC) [IMA][1440k][!]
- Star Wars- X-Wing - B-Wing v1.2 (1993)(LucasArts Entertainment Company LLC) [IMA][1440k][!]

They were installed upon each other successively in that order.
DOSC has copy protection patches, Sound Blaster Pro config, and auto start.
For some reason the GOG variant's default config made the game run very slowly, so I just replaced it with the same config as the main entry.

Also added the Collector's CD-ROM edition, installed from Redump-verified release:

- [Star Wars - X-Wing - Collector's CD-ROM (USA)](http://redump.org/disc/21705/)

This edition upgrades the game to the TIE Fighter engine among other changes, and generally seems to be treated as a separate game than the floppy version. So I think it's probably okay to have both in the main dat, similar to SimCity Enhanced CD-ROM.


tomojira:
Updates:
- Moved floppy version to variants
- Fixed the year for CD version
- Added MT-32 to floppy version, and General MIDI to CD version (according to Vogons Wiki, MT-32 is only properly supported by the floppy version)

televandalist:
These all seem to run fine. The only issue I had was that they won't launch without the DOSC unless you run the config first, but I think that should be okay?

Had zero issues with running them with the DOSC files in the directory. General MIDI and MT-32 options worked for each game as well 👍

PR #99, opened 2024-07-25
Add Star Wars - TIE Fighter + Defender of the Empire

tomojira:
Added TIE Fighter with its expansion, installed from 3.5" floppy disks matching with TGOD-recorded releases:

- [Tie Fighter (ID: 2729)](https://www.goodolddays.net/diskimages/id%2C2729/)
- [Tie Fighter: Defender of the Empire (ID: 2766)](https://www.goodolddays.net/diskimages/id%2C2766/)

DOSC has copy protection patch, Sound Blaster 16 config, and auto start.
GOG's default config didn't work well, so the variant uses the same config as the main entry.

Also added the Collector's CD-ROM edition, installed from Redump-verified release:

- [Star Wars - TIE Fighter - Collector's CD-ROM (USA)](http://redump.org/disc/25234/)

`TIE.CD` needs to be edited to `C:\`.

This edition upgrades the game engine, has 640x480 resolution, adds new missions, and generally seems to be treated like a separate game. So similar to X-Wing, I thought it would probably be okay to have both in the main dat.


tomojira:
Updates:
- Moved floppy version to misc variants
- Updated links
- Changed to WAITMODECHANGE
- Added MT-32 and General MIDI to floppy version, and General MIDI to CD version (according to Vogons Wiki, MT-32 is only properly supported by the floppy version)
- Added an official patch to the floppy version to fix the sound drivers for MT-32 and General MIDI:
https://www.patches-scrolls.de/patch/4117/7/51883/download

tomojira:
More updates:
- Collector's CD-ROM: Changed to minimal installation, which is just a few files and runs from the CD
- Added `cpu_max_year: 1998` or `cpu_year: 1994`. Collector's CD-ROM uses protected mode, while the floppy version seems to use real mode
- Added setup options

televandalist:
Tested these alongside https://github.com/PureDOS/DAT/pull/98

Copy/pasting in case you read this one first:

These all seem to run fine. The only issue I had was that they won't launch without the DOSC unless you run the config first, but I think that should be okay?

Had zero issues with running them with the DOSC files in the directory. General MIDI and MT-32 options worked for each game as well 👍

PR #101, opened 2024-07-26
Add Ignition

televandalist:
- Installed from Redump-verified disc: [Ignition (USA) (En,Es,It,Sv) (Rerelease)](http://redump.org/disc/10521/)
- DOSC mounts CD and auto-starts


televandalist:
Added GOG variant

PR #103, opened 2024-07-26
Add Contra/Gryzor

tomojira:
Added Gryzor, using this TGOD-verified release:

- [Gryzor (ID: 2669)](https://www.goodolddays.net/diskimages/id%2C2669/)

A terrible port of Contra/Gryzor with CGA graphics and no music. TGOD doesn't have Contra (USA), just Gryzor (Europe).
It runs from the disk with no installation.
DOSC has auto start.


tomojira:
Changed to the file set extracted from the disk, rather than the disk itself, to be more consistent with other entries.
I also learned that setting `video_card: generic_cga` improves the colours, and I reduced the game speed to something more playable with `cpu_year: 1981` (315 cycles / 8088 4.77 MHz speed). According to this page, this is the intended speed for the game:
https://contrapedia.wordpress.com/contra-conversions-dos/
To be honest it feels like it might be too slow, but 1982 (900 cycles) and 610 cycles feel too fast. No speed can really fix this game.

PR #104, opened 2024-07-27
Add Metal Gear

tomojira:
Added Metal Gear, extracted from TGOD-verified release:

- [Metal Gear (ID: 955)](https://www.goodolddays.net/diskimages/id%2C955/)

`INSTALL.BAT` gave an error, so I extracted all files with WinImage.

The game doesn't seem to have audio, and I can't see any options or config files to change it. According to some comments on this YouTube video, you can enable audio by setting oplmode to none, but that doesn't seem to be an option in DOS.YML.
https://www.youtube.com/watch?v=4aiM4tiEQCQ

> -PC speaker music is enabled by setting oplmode to none in the config file (or in DOSBox before running the game, but you have to do that each time). But yeah, not worth it.

> Setting PC Speaker as the audio source alone in DOSBox isn't enough to get music working in this game. You also need to set oplmode to none.

> The music works just fine for my version of the game, even though my DOSBox settings are oplmode=auto. I just tried changing it to opimode=none, the music seems to play exactly the same.

> Still, your way of getting the PC speaker music running works faster than my way Setting pcspeaker=true and oplmode=none is much easier than changing all of the other sound settings.

DOSC has copy protection patch and auto start. `run_input: (ENTER)` is used to select EGA video from a menu.
The copy protection removal isn't great - it still asks you for a code, but you can enter anything to get past it.


PureDOS:
DOS.YML offers to fully remove the sound blaster card with `sound_card: none`. Is this too much for this game? Like does it have a way to play sound effects via sound blaster but keep the music via PC speaker? If not then that's probably what we should do. Also some aggressive CPU speed limit seems to be needed for some parts of the game.

tomojira:
Oh, that works! I didn't understand that option at the time. And I highly doubt the game uses any AdLib/Sound Blaster.
`cpu_year: 1982` seems like an okay speed. 8088/8086 is the minimum but 1981 is pretty slow to start.

Other changes:
- Added WAITMODECHANGE
- Removed "Game will still ask for copy protection code, but you can enter anything to get past it" from comment as I realized since then that's common with copy protection removal
- Updated media link format

PR #106, opened 2024-07-27
Add Star Control

televandalist:
- Installed from Redump-verified CD-ROM: [Star Control Collection (USA, Europe)](http://redump.org/disc/20924/)
- DOSC contains IPS patch to bypass copy protection + auto-starts the game


PureDOS:
Should we leave out STARCON.COM (and STARCON.BAT which launches just that)? It seems to be an "official" crack which fully skips the copy protection but causes the title screen music to continue play long into the game which seems weird. Maybe the COM file was added for the CD-ROM version? Running that .COM prelauncher vs. running the .EXE directly results in the game's memory to shift around so it might be best to just keep one executable.

televandalist:
> Should we leave out STARCON.COM (and STARCON.BAT which launches just that)? It seems to be an "official" crack which fully skips the copy protection but causes the title screen music to continue play long into the game which seems weird. Maybe the COM file was added for the CD-ROM version? Running that .COM prelauncher vs. running the .EXE directly results in the game's memory to shift around so it might be best to just keep one executable.

Took care of everything here: https://github.com/PureDOS/DAT/pull/106/commits/f6044b49d2c7fb57bac6ad25bffccf4fa9a7133f

PR #107, opened 2024-07-27
Add Star Control II

televandalist:
- Installed from Redump-verified CD-ROM: [Star Control Collection (USA, Europe)](http://redump.org/disc/20924/)
- DOSC contains IPS patch to bypass copy protection, a .BAT file that lets the player choose between Star Control II and Super Melee, and auto-starts said .BAT file.


PureDOS:
This seems a bit weird, the CD-ROM version seems to already have a crack with the .COM file based pre-launchers. The patches you included in the DOSC (MELEE.COM.IPS, STARCON2.COM.ips, STARCON2.EXE.ips) don't seem to change that. The EXE (which is only patched for STARCON2 not MELEE) still requires the correct copy protection code and the COM's still skip it. Again like the first Star Control game, these .COM pre-launchers cause the game memory to shift so ideally we would just have one way to launch the game. Maybe the CD-ROM version isn't the best for us?

televandalist:
> This seems a bit weird, the CD-ROM version seems to already have a crack with the .COM file based pre-launchers. The patches you included in the DOSC (MELEE.COM.IPS, STARCON2.COM.ips, STARCON2.EXE.ips) don't seem to change that. The EXE (which is only patched for STARCON2 not MELEE) still requires the correct copy protection code and the COM's still skip it. Again like the first Star Control game, these .COM pre-launchers cause the game memory to shift so ideally we would just have one way to launch the game. Maybe the CD-ROM version isn't the best for us?

I got rid of the IPS patches, but I'm not sure how to solve the pre-launcher issue. The game crashes after the copy protection if I run the EXEs instead of the COMs. 

I tried going with [TGOD 1005](https://www.goodolddays.net/diskimages/id%2C1005/), but both cracks I found for it are pre-launchers themselves and don't actually modify the existing files. The attached image is something I found in the GOG version's `STARCON2.COM` and `MELEE.COM`, which are not present in the floppy install.

So yeah, not sure what the best option is here. 😓

PureDOS:
> Hacked from the crack made by Law & Order Party so it will work on either STARCON2.EXE, or MELEE.EXE

Pretty funny that that is shipped on multiple official CD-ROM releases of that game.
I also found some description of someone's effort back in the day to work around the copy protection:

> Many of the initial cracks for this game dis not completely crack the game. Accolade put in several ckecksums in the code to couase the game to freeze on the starmap when the checksums were not correctly dealt with. This crack takes care of these by rewriteing the copy protection code to do several things. The correct answers are heavily encrypted in memory. I rewrote the protection code to 1. de-crypt the correct answer 2. copy this into the buffer where what you typed is stored 3. re-encrypt the right answer 4. Jump into the rest of the protection code. This way, it is assured that the game thinks that you typed the right answer, because you did!

So I guess for this game there only are these pre-loader (TSR program) type patches to get around the copy protection so I think I'll just merge this now with them included in the DOSZ, as they are part of the official CD-ROM release so why not.

PR #109, opened 2024-07-27
Add Witchaven

10kmotorola:
Matches [Witchaven (Europe) (Rerelease)](http://redump.org/disc/2055/) with the official mouse patch applied (same patch GOG uses). I used this patch instead of the more recent one due to it including an updated `SETUP.EXE`. Maybe I should apply the most recent one in addition to this, as it updates the EXE and adds DWANGO support (I believe)? Now that I'm thinking about it that would make sense ...

Also adds a variant for EGwhaven, a collection of patches that GOG also included alongside the original game.


PureDOS:
This removes the intro movie from the CD, are we ok without that? I guess it would require including the whole 156 MB CD image, too. The only other way would be to include just the video file (INTRO.SMK, 41 MB) and have a custom batch file that plays that from C:\INTRO.SMK instead of `smackply D:\whaven\intro.smk /u1`.

10kmotorola:
I'm fine without the intro movie. Since I've been busy for the past few months I'm not entirely sure, what's the standard now?

PR #110, opened 2024-07-27
Add Mega Man

tomojira:
Added Mega Man, matching with TGOD-verified release:

- [Mega Man (ID: 463)](https://www.goodolddays.net/diskimages/id%2C463/)

It runs from the disk without an installation, and it has no music. It also runs way too fast, and it could be difficult to decide on a speed without knowing what was intended back then. But I found this TAS page which mentions `cpu_cycles: 500` is used as a standard on Speedrun.com, so I went with that.
https://tasvideos.org/6192S
https://www.speedrun.com/mega_man_dos

DOSC has auto start. `run_input: (WAIT:400)(RIGHT)(RIGHT)(DOWN)(DOWN)(DOWN)(DOWN)(RIGHT)(DOWN)(ENTER)` is used to select these options from a menu:

Graphics Card: VGA
Animation: On (default)
Masking: On (default)
Sound: On (default)
Joystick: Off (probably easier to just remap keyboard controls to a gamepad)

This YouTube video has an explanation of what the options do (seemingly no difference between EGA and VGA, seemingly no difference with the animation option, masking enables a foreground layer of graphics).
https://www.youtube.com/watch?v=Q7d0S9iNKUw


tomojira:
Changed to `cpu_year: 1981` (900 cycles) instead of using `cpu_cycles` directly. That makes it faster than the Speedrun.com rule, but I wasn't really sure about that anyway.

tomojira:
Changed to extracted file set to be more consistent with other entries.
Also changed the speed yet again to `cpu_cycles: 610`. This is around the middle of `cpu_year: 1981` (315 cycles) and `1982` (900 cycles), and similar to the game speed in 86Box with an 80286 at 6 MHz, which was used by the IBM PC/AT. This one forum post agrees this is a good speed for the game:
https://tasvideos.org/Forum/Topics/11865?CurrentPage=3&Highlight=290049#290049
Which doesn't mean much. But 1981 did feel too slow, and 1982 felt too fast.

PR #111, opened 2024-07-27
Add Mega Man 3

tomojira:
Added Mega Man 3, matching with TGOD-verified release:

- [Mega Man 3: The Robots are Revolting (ID: 896)](https://www.goodolddays.net/diskimages/id%2C896/)

Similar to the first Mega Man game, it runs from the disk without an installation, and has no music. It also runs too fast - I set `cpu_cycles: 1444` to slow it down, taken from the Speedrun.com rules.
https://www.speedrun.com/mega_man_3_dos?h=Any&rules=game
This is different than the first game, but `500` seems more painfully slow here, and I guess this is the only standard to go off.

DOSC has auto start. `run_input: (WAIT:400)(RIGHT)(DOWN)(DOWN)(DOWN)(DOWN)(RIGHT)(DOWN)(ENTER)(WAIT:12000)` is used to select these options from a menu:

Graphics Card: EGA (no VGA option like the first game)
Animation: On (default)
Masking: On (default)
Sound: On (default)
Joystick: Off

It takes a very long time to load at a black screen after choosing these options, which I mostly skipped through. After that you have to press space (fire) to get through the logos, which is pretty slow and unresponsive, so it may not be obvious what's going on.


tomojira:
Like the first Mega Man game, changed to `cpu_year: 1981` (900 cycles). That might actually feel a bit better in-game, but it's definitely sluggish getting started.

tomojira:
Changed to extracted file set to be more consistent with other entries.
Also updated to `cpu_year: 1982`, the new setting for 900 cycles. The game speed is different than the first Mega Man - 610 cycles is too slow with this one, but 900 feels fine.

PR #113, opened 2024-07-28
Add Boulder Dash 1 & 2

tomojira:
Added Boulder Dash, matching with this TGOD-recorded release:

- [Boulder Dash (ID: 2374)](https://www.goodolddays.net/diskimages/id%2C2374/)

This is a PC booter game that supports CGA and PCjr. The PCjr version has better graphics and sound, but seems buggy or perhaps not emulated as well? There are some odd graphics on the title screen, and it can't seem to get past the first level. So I went with CGA.

More information on the versions:
https://nerdlypleasures.blogspot.com/2016/08/boulder-dash-pc-speed-and-joysticks.html

DOSC has auto boot.


tomojira:
I suspect the version I initially added is modified, so I've replaced it with a DOS conversion recorded by TDC, and added Boulder Dash II as well. They're from the Super Boulder Dash compilation:
`Super Boulder Dash [DC] (1986)(Electronic Arts, Inc.) [Action, Strategy, Compilation]`

Of course these are basically hacks as well, but if that's the only option, I think this is the better way to go (and the only real way for Boulder Dash II). I like that TDC is more clear about what these are, and the PCjr mode seems to work better, it can get past the first level (though there's still odd graphics on the title screen). The Super Boulder Dash versions also add a Tandy mode as another option, which seems pretty much the same as PCjr. It looks and sounds much better than CGA anyway.

I've added these as variants without main entries, in case something gets done with the raw flux dumps and the pure PC booter disks can be added there. Not really sure about that though.

DOSC has PCjr mode and auto start.

PureDOS:
Thanks for doing all this research!

PR #115, opened 2024-07-28
Add Jill of the Jungle

televandalist:
- Installed from Redump-verified CD-ROM of [Jill of the Jungle (Europe) (Rev 1)](http://redump.org/disc/58954/)
- DOSC auto-starts


PureDOS:
The trilogy release (version 1.2 according to JILL.DOC) on the CD version has this weird startup of Episode 2 which needs inputting of multiple keys to get into the game while Episode 1 and 3 can be started by just accepting the auto detected settings by pressing ENTER.

I wonder if we should switch to the free game available on GOG... That one says version 1.0 in JILL.DOC but all 3 Episodes have the same startup config screen that can be skipped just by pressing ENTER. I think instead of using the episode selection launcher of the CD version we should offer the episode selection in the emulator startup menu. That way we could use `run_input` to skip past the configuration screen to make getting into the game more seamless. The only difference (besides Episode 2 config) is that there is a "ORDER INFO" menu item on the main menu on version 1.0 which seems to have been removed for the 1.2 trilogy release.

What do you think?

PureDOS:
Kept the CD version but removed "ORDER.FRM" so either CD-ROM release or the "The Complete Trilogy" floppy release can be used. I fixed the configuration of JILL2 by adding a pre-configured JILL2.CFG (which is just a copy of JILL1.CFG or JILL3.CFG) to the DOSC. I switched using of the trilogy launcher to using an episode selection in the start menu which then skips past the text menu setup screen directly to the title screen.
Also added GOG variant because it's available there for free.

PR #118, opened 2024-07-28
Add Command & Conquer - Red Alert + Counterstrike + The Aftermath

tomojira:
Added Red Alert with expansions, first installed from either of these 2 discs:

- [Command & Conquer - Red Alert (USA) (Allied Disc)](http://redump.org/disc/2196/)
- [Command & Conquer - Red Alert (USA) (Soviet Disc)](http://redump.org/disc/2197/)

Then these installed over it in this order:
- [Command & Conquer - Red Alert - Counterstrike (USA)](http://redump.org/disc/2198/)
- [Command & Conquer - Red Alert - The Aftermath (USA)](http://redump.org/disc/2199/)

This leaves it at v2.00EAM. There are later patches for v3.00 and v3.03, but these are only for the Windows version.
https://cnc.fandom.com/wiki/Red_Alert_patch_3.00_beta

Any one of the 4 discs is needed to start the game, and the expansion discs are needed to play their new missions. The Allied Disc and Soviet Disc have different `MAIN.MIX` files. I don't know what that implies, but I included both CD images in case they have different movies or something. Like the first C&C, this might be a mistake, and the file size is even larger here.


tomojira:
Updates:
- Split Red Alert, Counterstrike, and The Aftermath into parent/child
	- There are no patches for the DOS version of Red Alert, so on its own it's still v1.04, and the expansions patch it to v1.07E and v2.00EAM
- CD images converted with CHDtoOGG
- Windows-only files (except maybe the DLL files, not sure about those), ICO files, and PIF files removed
- Changed to WAITMODECHANGE
- I noticed that the expansions were creating 20 MB swap files, so I added `mem_size: 65536` to the DOS.YML. However the swap files are still created regardless
- Fixed config file so it doesn't skip the opening movies on the first time playing

PureDOS:
> I don't know what that implies, but I included both CD images in case they have different movies or something. Like the first C&C, this might be a mistake, and the file size is even larger here.

The C&C games have only the first one or two missions of both campaigns on both discs, afterwards the movies are only available on one disc, so it definitely needs both discs in the DOSZ. It will prompt you to swap the CD at that point if the wrong one is inserted.

Good catch on the 20 MB swap file. Though for me this did go away with "mem_size: 32768", did you start over with a clear save and still have it created?

Should games like this share the save files across the expansions? It feels a bit weird that after playing the main game then switching to an expansion and have that again start the main game's intro and intro level instead of going to the main menu (as would be the case if save files were shared).

tomojira:
Sorry, I had missed this reply!
Trying again I found that the swap file wasn't created, so I did make some mistake last time. In any case I reduced the mem_size to 32768.
Also added a shared .savename.

PR #119, opened 2024-07-29
Add Panzer General

tomojira:
Added Panzer General v1.2, installed from CD matching this Redump-verified release:

- [Panzer General (USA) (Rerelease)](http://redump.org/disc/80696/)

It gives an error unless the CD image is included. Though it may be possible to have an extracted install.

The game runs too fast, and it's difficult to determine how much the speed should be reduced. Ultimately I went with `cpu_cycles: 9000` just based on random forum post I found in here:
http://www.abandonia.com/vbullet/showthread.php?t=33780
The opening credits/movies seem okay with this setting. But it may take an expert on the game to decide on the ideal speed.


tomojira:
Changed to `cpu_year: 1990` (14000 cycles). This seems like a reasonable enough speed for the opening credits - maybe too fast, but 1989 (7800 cycles) might be too slow.

tomojira:
Updates:
- Change to WAITMODECHANGE
- Change to cpu_year
- Add General MIDI option (Sound Canvas)

After taking another look, I don't think it's possible to have a no-CD setup - it seems to be hardcoded, with no path to edit anywhere.

PR #120, opened 2024-07-29
Add Starflight

tomojira:
Added Starflight, matching with files extracted from:

- [Starflight (ID: 1595)](https://www.goodolddays.net/diskimages/id%2C1595/)

`HDINSTL.BAT` failed to install the game, so I just extracted everything from the disks instead.
There is a DOS.YML in the DOSZ with `cpu_cycles: 1500` and `video_card: generic_cga` to make the game playable - otherwise it can launch with glitchy graphics. This was taken from the PCGameWiki page.
https://www.pcgamingwiki.com/wiki/Starflight

DOSC has auto start, and Color TV/Composite mode selected from a menu with `run_input: (WAIT:300)3(ENTER)`.


tomojira:
Okay, I was too hasty submitting this. Investigating more, I found there are at least 5 variations in distributions.

1. Version 1, 1986, two 5.25" floppy disks. `HDINSTL.BAT` fails to install. `STRFLINS.BAT` mostly works, but fails to copy `STARB.COM` after swapping disks. There is one text file, `HDNOTES.TXT`.
TGOD: [Starflight (ID: 1595)](https://www.goodolddays.net/diskimages/id%2C1595/)
TDC: `Starflight (1986)(Electronic Arts, Inc.) [IMA][360k][!]`

2. Version 1, 1987, one 3.5" floppy disk. `HDINSTL.BAT` is functional, and there is no `STRFLINS.BAT`. Adds a few more files: `GOBACK.BAT`, `REMEMBER.BAT`, `THROW.BAT`, `YESNO.COM`. `HDNOTES.TXT` is different.
TGOD: [Starflight (ID: 1603)](https://www.goodolddays.net/diskimages/id%2C1603/)
TDC: `Starflight (1987)(Electronic Arts, Inc.) [IMA][720k][!]`

3. Version 2, 1987, two 5.25" floppy disks. `HDINSTL.BAT` is functional, including after swapping disks, and `STRFLINS.BAT` is back. `THROW.BAT` is gone and replaced with `THROWOUT.BAT`. `REMEMBER.BAT` is changed, and `XXXXXXXX.BAT` is added. There are two text files, `HDNOTESA.TXT` and `HDNOTESB.TXT`. `STARA.COM`, `STARB.COM`, and `STARFLT.COM` are updated with version 2.
TGOD: n/a
TDC: `Starflight (1987)(Electronic Arts, Inc.) [IMA][360k][!]`

4. Version 2, 1988, one 3.5" floppy disk. `HDINSTL.BAT` is functional, though it still asks you to swap disks, but this has no effect. Otherwise the same as the previous entry.
TGOD: n/a
TDC: `Starflight (1988)(Electronic Arts, Inc.) [IMA][720k][!]`

5. Version 2, GOG. Files are now split between 2 folders, `PLAY` and `START` (previous entries only had `PLAY`). `RESET.BAT` is added.
Unknown source - there is nothing else recorded by TGOD and TDC (except one bad dump on TGOD). Unless GOG themselves made up `RESET.BAT` for some reason - it doesn't do anything except copy files from `START` to `PLAY`.

A little bit of information on the versions:
https://www.vogons.org/viewtopic.php?t=40352
Version 2 adds EGA support, and there are a few gameplay differences described. It's mentioned that purists might prefer version 1.

So all that considered, I'll change the main entry to version 2, matching with both of these TDC entries:
`Starflight (1987)(Electronic Arts, Inc.) [IMA][360k][!]`
`Starflight (1988)(Electronic Arts, Inc.) [IMA][720k][!]`
Unfortunately this means no TGOD link, but it simply doesn't have version 2.

If purists might prefer version 1, I'll add that as a variant, matching with TGOD:
[Starflight (ID: 1603)](https://www.goodolddays.net/diskimages/id%2C1603/)
Instead of the ID 1595 I initially added, with its dysfunctional installer and fewer files.

And I'll keep the GOG variant for now, though honestly it is very thin, and could probably just be removed.

tomojira:
Version 1 does still need `video_card: generic_cga` in the DOSZ to be playable after selecting Color TV/Composite mode. However the `cpu_cycles: 1500` was unneeded so I removed that.
`run_input: 3(ENTER)` is used to select the mode.

Version 2 does not need a DOS.YML in the DOSZ with its EGA mode. (I completely missed that initially with the GOG variant)
`run_input: (WAIT:120)5(ENTER)(WAIT:120)` is used to select the mode.

tomojira:
On second thought, yeah I'll just remove the GOG variant. The only point of `RESET.BAT` and the duplicated files in the `START` folder is to restore the original files if they get overwritten. But that wouldn't happen in a DOSZ anyway, so there's no point in having a variant for this.

PureDOS:
It looks like the functionality of REMEMBER.BAT, GOBACK.BAT and THROWOUT.BAT requires the game to reside under C:\STARFLT\PLAY, then REMEMBER.BAT would copy the game to another directory and later one could go back to a remembered state by having GOBACK.BAT copy it back into C:\STARFLT\PLAY (or THROWOUT.BAT to delete the copied files). So we either remove these batch files (so a user would need to backup the emulator game save state if they desired) and move the game into C:\ (maybe only STARFLT.COM/STARA.COM/STARB.COM) or we have the game under C:\STARFLT\PLAY in the DOSZ.

tomojira:
I moved the files into `C:\STARFLT`, now `remember name` and `goback name` seem to work as intended. Also added a DOSC option for CGA composite mode, this has a slight advantage over EGA in showing animations on the title screen, and possibly other differences.

PR #122, opened 2024-08-02
Add Battle Arena Toshinden

televandalist:
- Installed using Redump-verified CD-ROM: [Battle Arena Toshinden (USA)](http://redump.org/disc/46394/), which was added to `Variants_Install.xml`.

**Initial Setup:**
- Recommended Installation (15 MEG)
- Install Directory: C:\TSD
- Digital Device: Sound Blaster AWE32 (Port 220, DMA 1, IRQ 7)
- MIDI Device: Sound Blaster Pro (MIDI Port 388)

**Misc. Options:**
- Music: CD Audio
- SVGA: Enabled
- 3D Acceleration: Auto-Detect (which ends up being NONE)

- DOSC configures all of the above, sets `cpu_year` to 1996 since the game ran too fast otherwise, mounts disc, and auto-starts.

**3DFX Variant**
- `SETUP.EXE` and `TSD.EXE` replaced with [v2.5 upgrade to support 3DFX Interactive Voodoo Graphics](https://archive.org/details/TSDUPGRD).
- DOSC is mostly the same as the unpatched version, but auto-starts a batch file specifically for 3DFX settings


PureDOS:
Should we merge the 2 versions? Is there any downside running the 2.5 update without 3dfx? It seems to run fine. 3DFX can then be a mode selectable on startup via a dosc config.

PR #123, opened 2024-08-02
Add Boulder Dash Construction Kit

tomojira:
Added Boulder Dash Construction Kit, matching with TGOD-verified 3.5" floppy disk:

- [Boulder Dash Construction Kit (ID: 2570)](https://www.goodolddays.net/diskimages/id%2C2570/)

It runs from the disk. Some levels are provided, not only the construction kit.

Unlike the first 2 games, there is no PCjr or Tandy mode, only CGA. `video_card: generic_cga` needs to be in DOS.YML or it will run slowly in-game.

DOSC has joystick disabled and auto start. `run_input: n(ENTER)` is used to choose no joystick from a menu, then press enter to start the animation and music on the title screen.


tomojira:
I decided to change this to the extracted file set, rather than just the disk itself, to be more consistent with other entries. I also just like being able to see all the files. Though in this case the disk also has some DOS system files, I still included everything.

PR #127, opened 2024-08-04
Add Jumpman

tomojira:
Added Jumpman DOS conversion from PC booter, matching with TDC-recorded file set:

`Jumpman [DC] (1983)(IBM) [Action]`

This is another one that has nothing on TGOD.

It has a PCjr mode with better graphics, and mostly better sound (except there is an annoying sound sometimes). However it runs way too fast, even at `cpu_year: 1980`. So I used `cpu_hz: 500`, as low as it goes, and that runs at a good speed.

DOSC has the default high score file, PCjr mode, no joystick (with `run_input: (WAIT:210)n`), and auto start.


tomojira:
Turns out even `cpu_hz: 500` is too fast for this game, comparing with the Jumpman Project, a fanmade modification designed to fix the speed. So I've added that as a variant.

DOSC has PCjr mode, no joystick (with `run_input: (WAIT:900)n`), and auto start.

tomojira:
Somehow I missed that TGOD does in fact have an entry for Jumpman... So I've changed to use a cracked PC booter disk that matches with this:

- [Jumpman (ID: 261)](https://www.goodolddays.net/diskimages/id%2C261/)

I've also realized that PCjr just runs too fast compared to CGA, even in 86Box with an 8088 either way. So I've decided to sacrifice the better graphics and use CGA for better gameplay. `cpu_cycles: 315` is now used to be accurate to an 8088.

DOSC has auto boot and `run_input: r(WAIT:70)n` to select RGB mode and no joystick.

PR #136, opened 2024-08-06
Add Powerslave

tomojira:
Added Powerslave, installed from CD matching with Redump-verified release:

- [Powerslave (USA)](http://redump.org/disc/24321/)

Prepare for an extremely slow installer.
CD image is needed to run the game, and provides CD audio.
"640x480" (actually seems to be 640x400) resolution needs to be set to fix flickering graphics on the title screen.

DOSC has Sound Blaster 16 config, Keyboard and Mouse controls, 640 screen width, and auto start.
A GOG variant is added as well.

EDIT: The flickering graphics can be an issue with Build Engine games when CPU cycles exceed 220,000:
Increasing the resolution fixed it on the title screen in this case, but I added `cpu_year: 1996` to prevent any more issues.


tomojira:
Updates:
- Changed to WAITMODECHANGE
- Changed to `cpu_max_year: 2000` - still works for fixing the flickering in other Build Engine games, while allowing better performance in-game

PR #137, opened 2024-08-06
Add Redneck Rampage & Suckin' Grits on Route 66

tomojira:
Added Redneck Rampage with Cuss Pack and expansion, installed from CDs matching with Redump:

- [Redneck Rampage (USA)](http://redump.org/disc/10509/)
- [Redneck Rampage - Cuss Pack (USA)](http://redump.org/disc/21890/)
- [Redneck Rampage - Suckin' Grits on Route 66 (USA)](http://redump.org/disc/20205/)

The Cuss Pack files need to be extracted/copied from the disc.
Suckin' Grits has a rerelease verified by Redump as well, but I found the executables didn't match. I went with the original release because its executables match with GOG.
When running Suckin' Grits, it will rename and backup some files, and create a new `RR.EXE`. It all seems to work fine, but you can see it doing stuff in a `FILEMODS.DBP` file.

DOSC has Sound Blaster 16 config, Cuss Pack enabled, and auto start.
There are also GOG variants.

EDIT: Flickering graphics can be a problem with Build Engine games when CPU cycles exceed 220,000:
Added `cpu_year: 1997` to prevent this issue.


televandalist:
The `<link type` lines should end with `</link>` instead of `</store>`.

tomojira:
I've learned that EXE files won't be able to run from the `.pure.zip` file with RA on hardcore mode. So the new `RR.EXE` that `ROUTE66.EXE` generates has been added to the DOSZ for Suckin' Grits. The game may need to be opened once to get that file to build the DOSZ.
Also removed `_ISDEL.EXE` after realizing it's an InstallShield file.

tomojira:
Updates:
- Change to WAITMODECHANGE
- Change to `cpu_max_year: 2000` - still fixes the flickering, while allowing better performance in-game
- Suckin' Grits developer changed to Sunstorm Interactive
- Removed `REDNECK.CFG` from GOG DOSZ
- Removed `_ISDEL.EXE` from GOG Suckin' Grits, and added `RR.EXE` (created after running `ROUTE66.EXE`)
- GOG config is now mostly (except for the Cuss Pack) the same as main config to reduce complexity

PR #138, opened 2024-08-06
Add Redneck Rampage Rides Again

tomojira:
Added Redneck Rampage Rides Again, installed from CD matching with Redump:

- [Redneck Rampage Rides Again - Arkansas (USA)](http://redump.org/disc/20193/)

(I don't think "Arkansas" is actually part of the title, Redump is just literal like that sometimes)
Unlike the first game, the CD music doesn't play automatically, and there doesn't seem to be an option to enable it. You have to go in the menu to play it every time.
The readme mentions another Cuss Pack, but it seems like it was never released.
https://www.gog.com/forum/redneck_rampage_series/rides_again_cuss_pack

DOSC has Sound Blaster 16 config and auto start. This case is unusual because the DOS console stuff appears _after_ the opening movie, so it isn't skipped.
GOG variant is added as well.

EDIT: Flickering graphics can be a problem with Build Engine games when CPU cycles exceed 220,000:
Added `cpu_year: 1998` to fix this.


tomojira:
Updates:
- Changed to WAITMODECHANGE
- Changed to `cpu_max_year: 2000` - still fixes the flickering, while allowing better performance in-game

PR #139, opened 2024-08-06
Add Zeliard

Greyfalken:
### Source Media

- Installed v2.0 from 2 3.5" floppy disks matching The Good Old Days ID: [2451](https://www.goodolddays.net/en/diskimages/?id=2451)
- Matches TDC Entry: `Zeliard v2.0 (1991)(Sierra On-Line, Inc.) [Action, Adventure]`

Removed the existing save game files `STEVEN.USR` and `ASS.USR` (lol) from the TGOD install.

### Installation Options

Configuration is contained in the DOSC instead of the DOSZ, so installation options do not matter. To play the game without the DOSC, users can run `INSTALL.EXE` to produce `RESOURCE.CFG`.

### Official Patches

### DOSC Contents

- Auto start
- Configuration file
- MT-32 option

### Installation Variants

Install disks matching TGOD ID 2451 have been added as an installation variant.

### Digital Store Variants

### Unofficial Patch Variants


Greyfalken:
Updated the DOSC to add `(WAITMODECHANGE)` and MT-32 support.

PR #146, opened 2024-08-09
Add Shadow Warrior & expansions

tomojira:
Added Shadow Warrior v1.2, installed from CD matching with Redump-verified release:

- [Shadow Warrior (USA)](http://redump.org/disc/2010/)

And its expansions, which are freeware:

Twin Dragon
https://www.dukeworld.com/2001-current/buildgames/sw/tc/tdragon/

Wanton Destruction
https://dukeworld.com/2001-current/apogee%20freeware/swwd.exe
(This is an InstallShield wizard that runs in Windows, but you can also open and extract it with 7-Zip)

Actually, Shadow Warrior itself is freeware from GOG, and it comes with the expansions too. So these are variants.
https://www.gog.com/en/game/shadow_warrior_complete

I didn't include the Build Editor during installation. The CD image is needed for music.
To fix flickering graphics, `cpu_year: 1997` is set.
DOSC has Sound Blaster 16 config, 640x480 resolution, and auto start. Wanton Destruction also has an input to select single player mode.


tomojira:
Updates:
- Changed to `cpu_max_year: 2000`. The flickering is still fixed when it's this high, while allowing for better performance in-game
- Changed to WAITMODECHANGE
- GOG config is now the same as main config for less complexity
- Titles changed to just "Twin Dragon" and "Wanton Destruction", since that's how they're presented in readmes/cover art

tomojira:
I see what you mean, it will automatically copy and rename `Wanton.dat` to a new `SW.EXE`. Added that directly to the file list instead and deleted the DOSC.
Also added "Shadow Warrior" back to the titles.

PureDOS:
I removed TEN now I don't think it serves a purpose and I think this can be merged. One thing I wanted to confirm, every time Twin Dragon is started, the Music option in the Sound Menu is automatically turned off, do you also get that? There seems to be no way to save it with music enabled, so I think that's just how it works.

tomojira:
I didn't notice that but yeah, I see it now. I've also seen it before with Redneck Rampage Rides Again, another Build Engine game with CD music. I never found a way to save the setting there either, so maybe it is intended.

PR #148, opened 2024-08-09
Add The Black Cauldron

Greyfalken:
### Source Media

- Installed v2.10 from a 3.5" floppy disk matching The Good Old Days ID: [950](https://www.goodolddays.net/en/diskimages/?id=950); also matches files installed from TGOD ID 1327
- Installed files also match TDC Entry: `Black Cauldron, The v2.10 [a1] (1988)(Sierra On-Line, Inc.) [Adventure]`

### Installation Options

No options during installation.

### Official Patches

### DOSC Contents

- CPU year set to 1986
- Auto start
- Tandy (3 voice) option via command line argument

### Installation Variants

Install disk matching TGOD ID 950 has been added as an installation variant.

### Digital Store Variants

### Unofficial Patch Variants


Greyfalken:
Updated DOSC to add `(WAITMODECHANGE)` and Tandy (3 voice) support.

Greyfalken:
I had a similar thought, but felt some users might have grown up playing the PC speaker version and maybe would want to play the version with sounds they remember? I don't know, personally I would rather play with the better-sounding version that I hadn't heard before, so...maybe you're right.

PR #151, opened 2024-08-10
Add Zelda Classic

tomojira:
Added Zelda Classic v1.90 and v2.10, the last DOS versions of this game, available from these links:

v1.90
https://web.archive.org/web/20120812031747/http://www.zeldaclassic.com/files/zc190sr1.exe
This is an installer that runs in Windows.

v2.10
https://web.archive.org/web/20120812031747/http://www.zeldaclassic.com/files/zc210d.zip
https://web.archive.org/web/20120812031747/http://www.zeldaclassic.com/files/zc210w.exe
The DOS version alone is missing many files, but they can be copied from the Windows version after installing it.

Both also need CWSDPMI v7:
https://sandmann.dotster.com/cwsdpmi

I added both v1.90 and v2.10 because as far as I can tell, Zelda Classic isn't great with backwards compatibility.

v1.90 quests:
https://www.zeldaclassic.com/resource-listing/?filter_4=1.90
https://www.purezc.net/index.php?page=quests&version=1.90&sort=title&sortdir=A-Z

v2.10 quests:
https://www.zeldaclassic.com/resource-listing/?filter_4=2.10
https://www.purezc.net/index.php?page=quests&version=2.10&sort=title&sortdir=A-Z

Note that according to this page, v1.92 is also for DOS:
https://www.zeldaclassic.com/download-archive/
But this doesn't appear to be true, it only contains Windows versions. There aren't as many quests available for v1.92 anyway.

Anyway these games have problems running, but these instructions worked for me:
https://www.purezc.net/forums/index.php?showtopic=72675&p=1027304
You need to run CWSDPMI with `cwsdpmi -x` followed by the game exe. A BAT file in the DOSC will do this. `mem_size: 65536` is also needed. Without these 2 options, the games will crash or run slowly.

`ag.cfg` also needs to be edited to `qst_dir = C:\` for v1.90, or `qst_dir=C:\` for v2.10. I included the exact required file in the DOSC's anyway.

DOSC also has auto start.


tomojira:
Updates:
- Upon testing more, I've realized v2.10 is unstable and often crashes. Not sure if this is solvable, but in case it's just inherent to this version, I moved it to the misc. variants and made v1.90 the primary entry
- The file path in `ag.cfg` wasn't needed, so I've replaced it with an empty file (v1.90 needs it to exist - though it can also be created with `agsetup.exe`, starting with an empty file is a bit more convenient)
- Realized v2.10 didn't have `agsetup.exe`, so I added it from v1.90 along with `AGsetup.txt`
- The `ag.cfg` in the DOSC now has the default Sound Blaster 16 config
- v2.10 now has the default `zc.sav` file moved to the DOSC

PureDOS:
What is the reason for having `zc.sav` in the DOSC? If it is integral to the game or at least nice to have to match the original experience, it should be in DOSZ. Otherwise I think it would be fine if we excluded it.

tomojira:
The only reason is because it came with the download to begin with. However all the game saves are just new games in several of the quests it comes with, so it can be left out without any issue.

tomojira:
Updates:
- Removed `zc.sav` from DOSC
- Changed to WAITMODECHANGE
- Added `cpu_max_year: 1999`
- Moved comments around in dat and changed to indirect links
- Renamed `ZELDA.BAT` to `PLAY.BAT` (just out of slight preference/consistency with other things I did)

PR #157, opened 2024-08-11
Add Lode Runner - The Legend Returns

tomojira:
Added Lode Runner: The Legend Returns, installed from 3.5" floppy disks matching with TGOD:

- [Lode Runner: The Legend Returns (ID: 2589)](https://www.goodolddays.net/diskimages/id%2C2589/)

ID 1681 and all the CD releases only have the Windows version (or only have a Windows installer, anyway).

The default FM MIDI option was chosen during installation instead of General MIDI.
Sound config is handled by copying and renaming files from the `AUDIO` folder, rather than a config file.
The DOSZ alone will fail to run the game until `SETSOUND.BAT` sets this up.

DOSC has Sound Blaster 16 config (using XOR patches) and auto start. There's also an MT-32 DOSC variant.

EDIT: Deleted `LR.BAT`, as it seems it might vary for people, and it's useless anyway.


tomojira:
According to this page, this game is not actually compatible with MT-32, even though it claims to be:
https://www.vogonswiki.com/index.php/List_of_MT-32-compatible_computer_games#Games_that_falsely_claim_MT-32_compatibility
Which makes sense, the music sounded pretty bad. I have removed the MT-32 DOSC.

tomojira:
New installation instructions:
```
Installed v1.0 from 3.5" floppy disks matching with TGOD ID: 2589, and also with TDC entry: Lode Runner- The Legend Returns v1.0 (1994)(Sierra On-Line, Inc.) [IMA][1440k]
Installation options:
	Directory: C:\SIERRA\LODERUNN (default)
	Sound: Sound Blaster 16, Port 220
	MIDI: Sound Blaster Pro - OPL3
	Device: FM MIDI device
Then make a separate install and choose:
	MIDI: Roland Sound Canvas
	Device: General MIDI device
This will give the LR16.PRS and LR256.PRS files found in the GM folder
```
These versions of the `LR16.PRS` and `LR256.PRS` files are needed for General MIDI music to work properly.

PR #158, opened 2024-08-11
Add Bubble Bobble also featuring Rainbow Islands

tomojira:
Added "Bubble Bobble also featuring Rainbow Islands", installed from CD matching with Redump:

- [Bubble Bobble also featuring Rainbow Islands (USA, Europe)](http://redump.org/disc/103801/)

CD image is needed to run the game.


tomojira:
Updates:
- Changed to WAITMODECHANGE
- Added `cpu_max_year: 1996`
- Didn't add General MIDI option because only SB gives music

PureDOS:
Adventure Island seems to play MIDI music (unlike Bubble Bobble), but it does not sound great to me so I don't think anyone would miss it.

tomojira:
Yeah, I'm okay with that.

televandalist:
FYI the comment field in the DAT doesn't mention that it's the Redump version

PR #159, opened 2024-08-11
Add Bust-A-Move 2

tomojira:
Added Bust-A-Move 2: Arcade Edition, installed from CD matching with Redump:

- [Bust-A-Move 2 - Arcade Edition (Europe)](http://redump.org/disc/56075/)

CD image is needed to play the game, and for audio.

The game will present many resolution options, but increasing it seems rather pointless, as in-game it will always appear in a small window.
Two files are affected by this option, `BM2SETUP.DAT` and `UNIVBE.DRV`. The game is playable when they're left out, but there will be some flickering graphics on the menu screen, so they're in the DOSC.
Another file, `INSTALL.LOG`, seemed safe to leave out.

DOSC has the default 320x240 VESA VBE (using an XOR patch for `UNIVBE.DRV`) and auto start.


tomojira:
Updates:
- CD image converted with CHDtoOGG
- Changed to WAITMODECHANGE
- Added `cpu_max_year: 1996`

PR #163, opened 2024-08-11
Add Shanghai II: Dragon's Eye

televandalist:
**Main:**
- Installed from Redump-verified CD-ROM: [Shanghai II - Dragon's Eye (USA)](http://redump.org/disc/67349/)

**Variants:**
- Above disc added to installation DAT

**DOSC:**
- Auto-start and config


televandalist:
Updated the DOSC to support MT-32

PR #171, opened 2024-08-12
Add Super C

tomojira:
Added Super C. This file set matches with TDC entry:

`Super C (1990)(Konami, Inc.) [Action]`

It also matches with an installation from TGOD [ID 2362](https://www.goodolddays.net/diskimages/id%2C2362/), except that it includes the copy protection removal launcher `LOAD.COM`. I tried to make a XOR patch to add this file instead, but it didn't work - it just created a file called `LOAD-COM.XOR`. So I've submitted this for now.

DOSC has EGA video, AdLib sound, and auto start. These options are chosen from a menu with `run_input: (DOWN)(DOWN)(RIGHT)(DOWN)(DOWN)(ENTER)(WAIT:375)`.


PureDOS:
LOAD.COM seems to be just 256 bytes so we can probably just add it to the DOSC. Then the real floppy can be used to make the DOSZ.

tomojira:
I don't know what problem I had with this before, but I've now added `LOAD.COM` as an XOR patch in the DOSC and removed it from the DOSZ. (I could also change it to the actual file in the DOSC if that's what you meant)
Another file, `GO.BAT`, was removed for mismatching between the floppy install and TDC. It is created again after running the game anyway.
This still matches with TDC, and now also matches with an install from 5.25" floppy disks matching with TGOD: [Super C (ID: 2362)](https://www.goodolddays.net/en/diskimages/?id=2362)
`cpu_year: 1989` was also added. I noticed that `GO.BAT` doesn't work at 1990 - it isn't used anyway, but 1989 works there so I guess it's a better speed for the game.

PR #178, opened 2024-08-13
Add Lands of Lore: The Throne of Chaos

Greyfalken:
## Lands of Lore: The Throne of Chaos (1993) (Westwood Studios)

Installed from a Redump-verified CD-ROM: [Lands of Lore - The Throne of Chaos (USA) (Rerelease)](http://redump.org/disc/71669/)

Installed the official 1.02 CD version patch found [here](https://archive.org/details/LOLCDPCH_ZIP). This is necessary to prevent the game from hanging after character selection in DOSBox.

DOSC sets the DOSBox sound card to `sbpro2` as the sound is mixed very low using the default setting. It also mounts the CD and auto starts. `LANDS.CFG` configures sound to SoundBlaster Pro and enables digitized samples.

There's a GOG version that I have as well, but it stores the CD data in a `GAME.DAT` file (which is just a renamed ISO) and I couldn't figure out how to get it to automount in DOSBox Pure.


Greyfalken:
Updated the DOSC files for both regular and GOG versions. Added options for sound (Sound Blaster/MT-32/General MIDI) and language (English/French/German). Added `WAITMODECHANGE` and an additional `WAIT` command for the MT-32 versions to skip the "initializing sound hardware" loading screen that only occurs when MT-32 is enabled.

PR #179, opened 2024-08-14
Add The Colonel's Bequest

Greyfalken:
### Source Media

- Installed v1.000.046 from 4 3.5" floppy disks matching The Good Old Days ID: [171](https://www.goodolddays.net/en/diskimages/?id=171); also matches files installed from TGOD ID 1126
- Installed files also match TDC Entry: `Colonel's Bequest, The v1.000.046 (1989)(Sierra On-Line, Inc.) [Adventure]`

### Installation Options

Configuration is contained in the DOSC instead of the DOSZ, so installation options do not matter. To play the game without the DOSC, users can run `INSTALL.EXE` to produce `RESOURCE.CFG`.

### Official Patches

### DOSC Contents

- CPU year set to 1989
- Auto start
- Configuration file
- MT-32 option
- Copy protection patch

Per the article found [here](https://www.benshoof.org/blog/case-cracked), one of the most common copy protection patches used with the game causes the RNG to break, resulting in numerous issues. The patch applied here should avoid breaking the RNG.

### Installation Variants

Install disks matching TGOD ID 171 have been added as an installation variant.

### Digital Store Variants

The GOG version matches the base, so no variant is provided.

The game is not available on Steam.

### Unofficial Patch Variants


Greyfalken:
Updated DOSC to add MT-32 support and a better implementation of the copy protection patch (that still retains the game's RNG) in XOR format. Also added some `WAIT` commands to skip some graphical anomalies at startup.

PR #181, opened 2024-08-14
Add The Secret of Monkey Island

tomojira:
Extracted files from the enhanced CD-ROM version, matching with Redump:

- [Secret of Monkey Island, The (USA) (En,Fr,De,Es,It)](http://redump.org/disc/11195/)

This one is odd because there's seemingly no installer, but it also refuses to run from the CD. You have to extract the files manually. Then it needs the CD inserted to run, and it has nice CD music as well.

Several languages are included, and can either be chosen from a launcher or run directly from their own executable. I used the launcher to start, but it could also be possible to jump straight into the English version, then have DOSC's for the other languages.

DOSC only has auto start.


tomojira:
Added VGA v1.1, installed from XL Kixx 3.5" floppy disks:

- [The Secret of Monkey Island (ID: 3012)](https://www.goodolddays.net/diskimages/id%2C3012/)

The VGA version can crash with a 486 + Sound Blaster, so this official patch is used to fix it:
http://static.dolimg.com/mh_netstorage/lucasfilm/patches/pc/Monkey.exe
It's a self-extracting exe that creates a new `MONKEY.EXE`, which simply replaces the one from the disk. More info here:
https://www.pcgamingwiki.com/wiki/The_Secret_of_Monkey_Island#Sound_Blaster_486_patch

DOSC has copy protection patch and auto start. There is also an MT-32 DOSC variant.

tomojira:
Updates:
- Changed the CD version so it just runs from the CD as intended, without files extracted (I don't know what problem I had with it before, but it's working fine now)
- Removed a few unnecessary installer files from the floppy version which aren't found in other releases
- Changed patch link so it's not a direct download

PR #186, opened 2024-08-15
Add Anvil of Dawn

Greyfalken:
## Anvil of Dawn (1995) (New World Computing)

Installed from a Redump-verified CD-ROM: [Anvil of Dawn (USA)](http://redump.org/disc/105138/).

Installation settings: Maximum Installation; Digital Device: SoundBlaster Pro (Port 220/DMA 1/IRQ 7); MIDI Device: Roland Sound Canvas (MIDI Port 330).

Installed the official 1.1b95 patch found [here](https://community.pcgamingwiki.com/files/file/182-anvil-of-dawn-patch/).

Copied all files from the `ANVIL\DATA` directory on the CD to the `DATA` directory on the hard drive. This allows the game to be run without the CD mounted.

DOSC auto starts and contains `HMISET.CFG` which configures sound.

Redump-verified CD-ROM added as an installation variant.

GOG version added as a digital variant. The CD is stored in a renamed ISO file called `Anvil.dat` but I copied the files out of it like above and it is no longer needed to run the game. The DOSC is the same as the main entry.


Greyfalken:
Made the following changes:
- Added `(WAITMODECHANGE)` to both versions.
- Changed the GOG version to not require extracting files from the CD image. There's really no reason to force the users to do that when the GOG version can be run as-is by changing the image filename from `Anvil.dat` to `Anvil.iso`. Especially since the GOG version doesn't match the base even with the files extracted.
- Added the default `HMISET.CFG` provided by GOG to the DOSZ of the GOG version. This allows the game to be run from the DOSZ without the DOSC. The DOSC still contains a configured version of `HMISET.CFG` that overrides the default.

PR #187, opened 2024-08-16
Add Might and Magic series

tomojira:
### Might and Magic I-III

These were installed from the Ultimate Might and Magic Archives CD:

- [Ultimate Might and Magic Archives, The (USA)](http://redump.org/disc/78286/)

It's not super convenient, the installer may need to run in Windows 9x - or you could use this program to extract the InstallShield CAB files:
https://github.com/twogood/unshield
But the benefit is these versions of I-III all match with GOG, and III has the copy protection removed (using a program called `Mm3.com`).

MM1 runs too fast and has speed reduced with `cpu_year: 1982` (900 cycles).
The EGA mode is broken and can cause crashes:
https://www.vogons.org/viewtopic.php?t=76376
It could be patched, but Tandy seems to look and sound the same anyway.

MM2 refuses to run when there's too much memory. A DOS.YML with `mem_doslimit: 512` is needed in the DOSZ to make it playable. (GOG reduces the conventional memory by only 4 kb using `loadfix -4 mm2.exe`. But I went down to 512 kb just because that's a solid standard number)

MM1's `ROSTER.DTA` and MM2's `ROSTER.DAT` contain premade characters, and are modified when you create new characters, but need to exist for the games to run. They could be replaced with 0 byte files, but I found that this caused bugs in MM2 (NPCs wouldn't appear). So I kept the default files in the DOSZ for both MM1 and MM2 to be safe.

MM1 DOSC has Tandy mode, game speed reduced, and auto start.
MM2 DOSC has MCGA/VGA mode and auto start.
MM3 DOSC has Sound Blaster config and auto start. MM3 also has an MT-32 DOSC variant.

### Might and Magic IV-V

Standalone Darkside of Xeen and non-voiced World of Xeen can also be installed from the Ultimate Might and Magic Archives - but not standalone Clouds of Xeen. All 3 can be installed from the Might and Magic Trilogy CD:

- [Might and Magic Trilogy (USA, Europe)](http://redump.org/disc/53003/)

It also has III, but not with `Mm3.com`.

The fully voiced CD-ROM version of World of Xeen can be installed from CDs matching with these:

- [Might and Magic - World of Xeen (USA) (Disc 1)](http://redump.org/disc/20403/)
- [Might and Magic - World of Xeen (USA) (Disc 2)](http://redump.org/disc/20404/)

The 3 options to install every file were chosen.
This version won't run from `C:\`, the files need to be in `C:\WORLD`.

GOG's World of Xeen does not match, so it's added a variant. GOG doesn't have standalone Clouds of Xeen.

Clouds of Xeen, Darkside of Xeen, and World of Xeen DOSCs have copy protection patches, Sound Blaster config, and auto start. They all have MT-32 DOSC variants.
Voiced World of Xeen doesn't need copy protection removal.

### Swords of Xeen

This was installed from the Might and Magic Trilogy CD, and also matches with GOG and with the Ultimate Might and Magic Archives CD.

DOSC has Sound Blaster config and auto start. And once again, there is an MT-32 DOSC variant.

EDIT: The games may be unstable with MT-32 unfortunately:
https://www.gog.com/forum/might_and_magic_series/world_of_xeen_sound_effects_issues/post3
Just noting this for now.


tomojira:
Updates:

- Changed III-V to `cpu_year: 1989` (7800 cycles). This is a somewhat more reasonable speed for the game intros and enemy animations, etc. and it fixes an issue with a disappearing cursor in IV-V (e.g. at the top of the character status screen). The tradeoff is it causes the cursor to flicker often, but unfortunately it doesn't seem possible to completely solve this. It was a problem on original hardware too:
https://github.com/exoscoriae/eXoDOS/issues/4948
https://www.gog.com/forum/might_and_magic_series/technical_troubles_invisible_mouse_in_might_and_magic_4/post8
I wasn't sure which direction to choose, but ultimately thought the fast speed would be more of a problem. You can get used to the flickering cursor quickly, and it's also possible to completely play with a keyboard and ignore the mouse.

- Fixed most volume issues in III-V by changing to `sound_card: sbpro2`, and also changing the config files from Sound Blaster 16 to Sound Blaster Pro when applicable. The MT-32 variants could still use improvement, especially Clouds of Xeen, but that can wait until DOS.YML options are added.

- Changed to XOR patches for IV-V (and the ones for Darkside/World of Xeen now actually work, I had somehow messed up the patches before)

- Changed to the new `run_input: (WAITMODECHANGE)`

tomojira:
Updates:
- World of Xeen audio tracks converted with CHDtoOGG
- Removed an extra file `BOSS.MM4` from GOG's floppy variant

tomojira:
More updates:
- Changed World of Xeen CD to minimal install size
- Changed MM1 to WAITMODECHANGE (apparently forgot to do this before)
- Added `cpu_year: 1989` to MM2
- Add General MIDI option to all the Xeen games
- Renamed DOSC options

tomojira:
Updates:

- Changed the titles of Clouds of Xeen and Darkside of Xeen to be numbered IV and V. These titles are commonly known and officially used by the [Ultimate Might and Magic Archives](https://www.mobygames.com/game/3161/ultimate-might-and-magic-archives/cover/group-5525/cover-10297/), so I think this might be more convenient. MM1's title was also edited a bit
- MM2 can import characters from MM1, so they now use a shared save. Files were moved into `C:\MM1` and `C:\MM2`
- The DOS.YML in the DOSZ for MM2 now has embedded data
- Links added, comments edited, etc.

unverified/PureDOSDAT_Unverified.xml
```diff
@@ -21954,6 +21954,462 @@
...
+		<developer>New World Computing</developer>
+		<comment>
+			Installed from the Ultimate Might and Magic Archives CD, matching with Redump: Ultimate Might and Magic Archives, The (USA)
+			Also matches with GOG.
+			Needs DOS.YML with mem_doslimit: 512
+		</comment>
+		<comment_dosc>MCGA/VGA mode and auto start</comment_dosc>
+		<link type="media">http://redump.org/disc/78286/</link>
+		<link type="store">https://www.gog.com/en/game/might_and_magic_6_limited_edition</link>
+		<rom name="DOS.YML" data="bWVtX2Rvc2xpbWl0OiA1MTINCg==" size="19" crc="153c6558" md5="2b2a6213b5682369e893bbb20b3364e9" sha1="6986030ecb1f3ced72d68f3fe7466d47769891e6" data="bWVtX2Rvc2xpbWl0OiA1MTINCg=="/>
```
jumphil, line 22272: The data field is duplicate, which causes an error when loading the DAT in RomVault

tomojira: Sorry about that - I've fixed it in this PR here
https://github.com/PureDOS/DAT/pull/460

PR #193, opened 2024-08-19
Add Heroes of Might and Magic I & II

tomojira:
### Heroes of Might and Magic

Installed from the original CD, matching with Redump:

- [Heroes of Might and Magic (USA)](http://redump.org/disc/83264/)

Then installed patch v1.2:
https://www.moddb.com/games/heroes-of-might-and-magic-a-strategic-quest/downloads/heroes-of-might-magic-v12-dos-patch

The game needs `HEROES.CFG` or it will freeze, but unfortunately this file often changes based on how the game was installed.
These settings were used during installation, but if this doesn't work out, the exact file is provided in the DOSC.

```
CPU Detected: 1349 Mhz 486 (can be consistent when cpu_year: 1995 is set)
Install Path: C:\HEROES
Sound Card: Sound Blaster 16
Music: 16 Bit Stereo
Video Card: VESA Compatible
CD-ROM Letter: D
Modem Options: None
```

The Heroes of Might and Magic Compendium CD has a later version, v1.3, but it also has lower quality music, and there is no standalone patch for v1.3 or any info on changes in it.
https://cloakedthargoid.wordpress.com/pc-games-collections/homm-compendium/
https://www.celestialheavens.com/forum/topic/11798

GOG's version says v1.2, but it doesn't match with any old version I checked. `homm1.gog` was renamed to `homm1.iso` so it would auto-mount.

Besides the required config file, DOSC has auto start.

### Heroes of Might and Magic II

Installed from the original CD, matching with Redump:

- [Heroes of Might and Magic II - The Succession Wars (USA, Europe)](http://redump.org/disc/53002/)

And patched to v1.3:
https://www.moddb.com/games/heroes-of-might-and-magic-ii-the-succession-wars/downloads/heroes-ii-the-succession-wars-v13-dos-patch

The installed The Price of Loyalty:

- [Heroes of Might and Magic II - The Price of Loyalty (USA)](http://redump.org/disc/68929/)

And patched to v2.1:
https://www.moddb.com/games/heroes-of-might-and-magic-ii-the-succession-wars/downloads/heroes-ii-the-price-of-loyalty-v21-dos-patch

This is quite slow and painful, but the complete Gold edition is only for Windows, and v1.3 updates files that the expansion and v2.1 don't. The end result almost matches with GOG's Gold edition, except for the CD image.


tomojira:
Updates:
- Succession Wars v1.3 and Price of Loyalty v2.1 are now split into parent/child
- Both CD images for HoMM2 are included now (some of the soundtrack is different)
- HoMM2 CD images converted with CHDtoOGG
- GOG HoMM2 config is now the same as the main config to reduce complexity
- GOG HoMM2 retitled to just The Price of Loyalty

tomojira:
More updates:
- Added `cpu_max_year: 1997` to HoMM1, `1998` to HoMM2
- Added General MIDI option to HoMM2

PR #196, opened 2024-08-20
Add The Immortal

tomojira:
### The Immortal

Installed from 5.25" floppy disks, matching with TGOD:

- [The Immortal (ID: 3172)](https://www.goodolddays.net/diskimages/id%2C3172/)

This also matches with GOG.

DOSC has copy protection patch, keyboard controls, and auto start. There is also an MT-32 DOSC variant.


televandalist:
I started looking into doing some MT-32 variants and checked this out as an example. However, it seems like the keyboard doesn't work? I'm stuck at the new game screen.

tomojira:
For some reason The Immortal only uses joystick controls by default. To switch to keyboard, you have to use a command line parameter, `immortal keyboard` (and it's still a little weird in how it works, you have to hold a direction while you select a menu option). MT-32 is also a parameter, so both would be `immortal keyboard roland`.
Actually I forgot this PR is outdated now - I have to update it to one merged DOSC so it works with the new start menu.

PR #197, opened 2024-08-21
Add Wizardry series

tomojira:
All installed from the Ultimate Wizardry Archives CD, matching with Redump:

- [Ultimate Wizardry Archives, The (USA)](http://redump.org/disc/80025/)

The installer may need Windows 9x, but all files can be copied straight from the disc as well.
All games have copy protection removed.

### Wizardry I-V

These have `video_card: generic_cga` to give a better colour palette. They also have game speed reduced with `cpu_year: 1982`. This may still feel a little too fast, and 1981 might be more authentic, but 1982 is nice for reducing the long load times and input lag.
One problem is IV's battle messages are too fast, but there is an in-game option to control their speed by pressing T.

I-III are the 1987 re-release versions of the games.
All the games have premade characters included as well, which is especially helpful for II and III since they can't create characters, and originally had to transfer characters from I.

### Wizardry: Bane of the Cosmic Forge

Game speed is reduced with `cpu_year: 1985`.
`sound_card: none` is used to enable PC speaker sound, which is slightly better than the OPL sound, due to the weird sound engine. It's bad either way, with lots of popping, but that's normal, and happened on original hardware too.
https://github.com/dosbox-staging/dosbox-staging/issues/3642

`SCENARIO.HDR` can vary with options, and `PCFILE.DBS` can vary with characters created, but both are needed for the game to run. The defaults are always on the disc anyway.

### Wizardry: Crusaders of the Dark Savant

Game speed is reduced with `cpu_year: 1989`. At 1990 or higher cycles, the game freezes at the copy protection entry.
The sound setting is left at the default PC speaker, for the same reason as VI. Music is set to Sound Blaster.
The default `SCENARIO.HDR` and `PCFILE.DBS` are included in the DOSZ, following along with VI.

All DOSCs have auto start of course.


tomojira:
Updates:

- Changed titles to be numbered like Wizardry II, etc. These titles are commonly known and they were officially used by the [Ultimate Wizardry Archives](https://www.mobygames.com/game/1367/the-ultimate-wizardry-archives/cover/group-33280/cover-90163/), so I think this might be more convenient
- Edited the years again. I believe these versions of Wizardry I & II are remakes released in 1987, after III in 1986. There were earlier versions before 1986 with more primitive graphics
- Wizardry I, II, III, and V use the shared save feature (IV isn't involved because you play as a predetermined character in that one). The files weren't moved into subfolders because they have to be together. After Wizardry I saves `SAVE1.DSK`, Wizardry II can transfer characters from it into `SAVE2.DSK` at Edge of Town -> Utilities -> Move Characters. III can do the same from I or II, and V can transfer characters from any of the 3. I itself also has the option, but I guess that's just for moving from I to I. You can't transfer backwards. There may be messages about making/swapping disks, but you don't have to do any of that, these 1998 DOS conversions automatically handle it
- VII can also transfer characters from VI, so the shared save feature is used there too. Files were moved into `C:\BANE` and `C:\DSAVANT`
- MT-32 option added to VII
- Media links added, comments edited, etc.

PR #201, opened 2024-08-30
Add Circuit's Edge

Greyfalken:
### Source Media

- Installed from 3 5.25" floppy disks matching The Good Old Days ID: [2685](https://www.goodolddays.net/en/diskimages/?id=2685)
- Installed files match TDC Entry: `Circuit's Edge (1990)(Infocom, Inc.) [Adventure]`; also match files installed from TDC Entry: `Circuit's Edge (1990)(Infocom, Inc.) [IMA][360k][!]`

### Installation Options

No options during installation.

### Official Patches

### DOSC Contents

- CPU year set to 1983 (this slows down animations like the main character blinking and interior location view movement to a reasonable speed)
- Auto start
- MT-32 sound option

Sound configuration is set via the `run_input` command at startup.

### Installation Variants

Install disks matching TGOD ID 2685 have been added as an installation variant.

### Digital Store Variants

### Unofficial Patch Variants


Greyfalken:
Updated DOSC to add MT-32 support.

Note: This game rewrites the config file at every startup, so once a sound type is selected in the PureDOS menu the only way to switch later is to remove the config file from the pure.zip created for the game.

PR #214, opened 2024-09-06
Add Ancient Domains of Mystery

tomojira:
### Ancient Domains of Mystery

Freeware v3.3.3, downloaded from here:

https://www.adom.de/home/downloads.html

DOSC has auto start.


tomojira:
Updates:
- Moved to Freeware dat
- Added `cpu_year: 1996`
- Changed to indirect link

PR #218, opened 2024-09-07
Add William Shatner's TekWar

tomojira:
### William Shatner's TekWar

Installed from CD matching with Redump:

- [William Shatner's Tekwar (USA)](http://redump.org/disc/21229/)

CD image is needed to run the game. Audio tracks were converted with CHDtoOGG.

The cutscenes stutter and slow down with Sound Blaster 16, so Sound Blaster Pro is used. Unfortunately this causes constant clicking in the cutscenes, but it's the better trade-off.
https://www.vogons.org/viewtopic.php?p=264100

I didn't notice any flickering graphics like other Build Engine games, and even going up to `cpu_max_year: 1998` still had the frame rate dropping, so I left it unlimited for better performance.

DOSC has Sound Blaster Pro config, 640x480 resolution, and auto start.


tomojira:
Updates:
- Re-converted audio tracks with CHDtoOGG v1.1
- Added `cpu_max_year: 2000` for consistency with other Build Engine games
- Renamed to "Capstone Software" instead of just "Capstone"

PR #219, opened 2024-09-07
Add Callahan's Crosstime Saloon

Greyfalken:
### Source Media

- Installed from Redump-verified CD-ROM: [Callahan's Crosstime Saloon (USA)](http://redump.org/disc/69449/)

### Installation Options

- Optimum Performance Installation

Sound configuration is contained in the DOSC instead of the DOSZ, so sound options do not matter during installation.

After install, copy all `*.Q` and `*.VOC` files from the CD to the install directory. Then make the following changes to `LEGEND.INI`:
- Change line 3 to `QDATA=C:\`
- Change line 4 to `PICDATA=C:\`
- Change line 5 to `VOCDATA=C:\`
- Change line 6 to `SAVEDATA=C:\`

These changes allow the game to be run without the CD mounted.

The game will run from the DOSZ without the DOSC but will not have sound. To enable sound, users can run `SETSOUND.EXE` to produce `MDI.INI` and `DIG.INI`.

### Official Patches

### DOSC Contents

- CPU max year set to 1997
- Auto start
- Configuration files

### Installation Variants

The Redump-verified CD-ROM has been added as an installation variant.

### Digital Store Variants

### Unofficial Patch Variants


Greyfalken:
And I just remembered to check the "Games that falsely claim MT-32 compatibility" list, so I've updated the DOSC again to remove MT-32 support

Greyfalken:
Moved the file `LEGEND.INI` from the DOSC to the DOSZ to allow the DOSZ to run standalone.

PR #222, opened 2024-09-08
Add Rayman games

tomojira:
### Rayman

Installed from CD matching with Redump:

- [Rayman (USA) (En,Fr,De)](http://redump.org/disc/57505/)

English language and SB16 sound were chosen during installation. The resulting `RAYMAN.CFG` is included in the DOSZ because the sound can't be set again without reinstalling the game.

I found that the installer crashed sometimes, and even the game itself crashed once after completing a level. My random guess was to set `cpu_year: 1995` in case the speed caused it, and I haven't seen it crash again since, but can't say for sure.

DOSC has English language, Sound Blaster 16 config, and auto start, plus options for French and German.

### Rayman Designer

Files copied from CD matching with Redump:

- [Rayman Designer (Europe) (En,Fr,De)](http://redump.org/disc/36177/)

There is an installer but it needs Windows.
The designer itself is also Windows-only, but it comes with 24 new levels that play in DOS.

I found this even more unstable than vanilla Rayman, but again used `cpu_year: 1995`. My best guess from looking things up is high cycles or dynamic cycles might cause problems.

DOSC has Sound Blaster 16 config and auto start, plus options for French and German.

### Rayman by His Fans

Files copied from CD matching with Redump:

- [Rayman Forever (USA)](http://redump.org/disc/45463/)

Not many English releases available for this level pack, and unfortunately this one messes up the music - specifically Track 8, 13, and 14. So the OGG files for those tracks are copied over (and renamed) from `Rayman (USA) (En,Fr,De)`.

DOSC has Sound Blaster 16 config and auto start. No language options this time.

### Rayman 60 Levels

Files copied from CD matching with Redump:

- [Rayman Compilation (Europe)](http://redump.org/disc/55659/)

This is one of the few releases of 60 Levels in English.

DOSC has Sound Blaster 16 config and auto start, plus a French option.

CD images are needed to run all games. Audio tracks were converted using CHDtoOGG.


tomojira:
Added GOG variants for Rayman, Rayman Designer, and Rayman by His Fans.
- Because they're distributed together using the same CD image, I've set these up as parent/child
- Rayman Designer includes French and German, but the other 2 are only in English
- GOG still has the problems with Track 8, 13, and 14, and is missing the intro for base Rayman
- GOG doesn't have Rayman 60 Levels

tomojira:
I tested the minimal install size for Rayman, but found that it added a loading screen in between levels, so I think maximum is best here. It has to be "Extended install with video" because the opening movie only seems to play when it's on the hard drive, not on the CD.

I also realized that `RAYMAN.CFG` varies depending on the order you do things in the installer. So I changed it to a version generated by this exact order, which is more straightforward than the order I did before:

```
Installed from CD.
Installation options, in order:
	English language
	Extended install with video
	Directory: C:\RAYMAN
	Sound: SB16, IRQ 7, Port 220, DMA 5
Exact RAYMAN.CFG file is in the DOSC
```

PR #223, opened 2024-09-09
Add Albion

Greyfalken:
### Source Media

- Installed from Redump-verified CD-ROM: [Albion (USA)](http://redump.org/disc/107035/)

### Installation Options

Configuration is contained in the DOSC instead of the DOSZ, so installation options do not matter. The game will run from the DOSZ without the DOSC but will not have sound. To enable sound, users can run `SETUP.EXE` to produce `MDI.INI` and `DIG.INI`.

### Official Patches

### DOSC Contents

- CPU max year set to 1995
- Auto start
- Configuration files

### Installation Variants

The CD is required to play, so no installation variant is provided.

### Digital Store Variants

The GOG version is provided as a variant. The DOSC matches the base.

### Unofficial Patch Variants


Greyfalken:
Updated both DOSCs to add `WAITMODECHANGE`. Did not add MT-32 support as the game appears on the "Games that falsely claim MT-32 compatibility" list.

Greyfalken:
Moved the file `SETUP.INI` from the DOSC to the DOSZ in both the CD-ROM and GOG versions to allow the DOSZs to run standalone.

PR #226, opened 2024-09-10
Add Gabriel Knight: Sins of the Fathers

tomojira:
### Gabriel Knight - Floppy Version

Installed v1.0 from 3.5" floppy disks matching with TGOD:

- [Gabriel Knight: Sins of the Fathers (ID: 1564)](https://www.goodolddays.net/diskimages/id%2C1564/)

Then patched to v1.0b:
https://web.archive.org/web/20240117182716/http://www.sierrahelp.com/Files/Patches/GabrielKnight/GK1Floppy(GKPAT10B).zip

There is no copy protection.

Game speed was reduced to try to avoid these bugs:
https://www.pcgamingwiki.com/wiki/Gabriel_Knight:_Sins_of_the_Fathers#Issues_fixed
It says under 7000 cycles, but I went with `cpu_max_year: 1992` (27,000 cycles) to avoid slowing down the game too much, hopefully this works out fine.
https://steamcommunity.com/app/495700/discussions/0/350540973999715439/

DOSC has Sound Blaster Pro and auto start, plus options for MT-32 and General MIDI.

Previously I've been adding both floppy and CD versions in the main dat when there's some differences between them, but the more I see, the more I realize this isn't a good idea. So this is a variant instead.

### Gabriel Knight - CD Version

Installed v1.1 from CD matching with Redump:

- [Gabriel Knight - Sins of the Fathers (USA, Europe) (Rerelease)](http://redump.org/disc/1767/)

Maximum installation with these options chosen:

```
Graphics: VESA
Music: Sound Blaster Pro (default)
Audio: Sound Blaster (default)
Directory: C:\SIERRA\GKCD (default)
```

`RESOURCE.CFG` then needs to be in the DOSZ. The exact file is provided in the DOSC.

It might be possible to have a no-CD setup, but the config file doesn't have a path to the AVI movie files that can be modified. So the CD image is included to be safe.
I haven't added a GOG variant for the same reason, it might rely on ScummVM.

`cpu_max_year: 1992` is used similarly to the floppy version.

DOSC has Sound Blaster Pro, VESA 640x480, and auto start, plus options for MT-32 and General MIDI.


tomojira:
Changed to a no-CD setup. (I think it's probably fine without the AVI folder, SEQ seems to be used for in-game movies)

From the dat comment:
```
Installation options:
	Graphics: VESA
	Music: Sound Blaster Pro
	Audio: Sound Blaster (default)
	Directory: C:\SIERRA\GKCD (default)
	Full installation size
	Yes to copy high resolution art
Then manually copy the AUD, PATCHES, SEQ, and SFX folders from the disc.
In RESOURCE.CFG:
	Delete line 5 (directory = \SIERRA\GKCD)
And edit these lines like this (no spaces at the end of each):
	audio=C:;C:\sfx;C:\patches;C:\aud
	sync=C:\aud
	movieDir=C:\seq
	patchDir=C:;C:\patches
RESOURCE.CFG needs to be in the DOSZ for the game to run.
Almost matches with GOG, except for RESOURCE.CFG (grab it from the DOSC)
```

Other updates:
- Changed to `cpu_year: 1991`, hopefully more likely to solve the timing issues people talk about
- WAIT time increased a bit

PR #255, opened 2024-09-24
Add The Incredible Machine

tomojira:
### The Even More Incredible Machine

- Files copied from CD matching with Redump: [Even More! Incredible Machine, The (USA)](http://redump.org/disc/8146/)
- CD image converted with CHDtoOGG using -n option
- `cpu_max_year: 1991` is used to slow down the game a bit, and to make the GOG version playable at all (1992 and 1993 give a black screen)
- DOSC has Sound Blaster config and auto start


tomojira:
Updates:
- Added the original Incredible Machine for those who might prefer it. Files were copied from a 3.5" floppy disk matching with TGOD: [The Incredible Machine (ID: 821)](https://www.goodolddays.net/en/diskimages/?id=821)
	- The game won't run at the default CPU cycles - `cpu_year: 1991` works
	- DOSC has copy protection patch, Sound Blaster config, and auto start, with an option for MT-32
	- General MIDI doesn't work
- Changed TEMIM from `cpu_max_year` to `cpu_year: 1991`
- Tried MT-32 and General MIDI options for TEMIM, but they didn't do anything - only the CD music plays
	- They do work for the GOG variant, so added them there
- Also tried the normal installation method for TEMIM, but found that the setup program doesn't work properly, it will try to reinstall the game. So manually copying all the files from the CD seems to be best

PR #256, opened 2024-09-24
Add The Incredible Machine 2

tomojira:
### The Incredible Machine 2

- Installed from CD matching with Redump: [Incredible Machine 2, The (USA) (En,Fr,De,Es)](http://redump.org/disc/64891/)
	- Full installation, English language chosen
- CD image converted with CHDtoOGG
- DOSC has Sound Blaster config and auto start, plus options for French, German, and Spanish


tomojira:
Updates:
- The French, German, and Spanish language files are now DAT'd, with the DOSC using XOR patches to move them around
- Added `cpu_year: 1994`
- Tried an installation with all files copied from the disc, and no data track, but voice clips won't play when you do this. Also tried the minimal install size, but the setup program wouldn't work properly, it would try to reinstall the game. So I think maximum size is best here
- Tried MT-32 and General MIDI options, but they don't do anything, it always plays the CD music

tomojira:
- Added a floppy variant, installed from 3.5" floppy disks matching with TGOD: [The Incredible Machine 2 (ID: 1015)](https://www.goodolddays.net/en/diskimages/?id=1015)
- English, French, German, and Spanish versions were all installed to get all the files
- DOSC has English language, Sound Blaster config, and auto start, with options for MT-32, French, German, and Spanish
- There is General MIDI but it doesn't seem to work properly without unofficial patches

PR #264, opened 2024-09-26
Add MT-32 support to Out of This World 

televandalist:


PureDOS:
For this to work there needs to be a configured CONFIG.DAT file with MT-32 enabled in the `[MT-32 Sound]` directory.
But this game with MT-32 only play sound effects, is it even worth offering this as an option?

televandalist:
> For this to work there needs to be a configured CONFIG.DAT file with MT-32 enabled in the `[MT-32 Sound]` directory. But this game with MT-32 only play sound effects, is it even worth offering this as an option?

I'd say yeah... it's only sound effects, but there are more of them with it enabled.

```
Out of this World for DOS supports Sound Blaster, Adlib, Pro Audio Spectrum, PC Speaker and Disney Sound Source, all at either 10 kHz or 5 kHz.  It also supports the Roland LAPC-I and CM-32L (and compatibles with a Roland MPU-401 interface), but not the MT-32 or MT-100.     It only uses the 33 extra sound effects found on the LAPC-I and CM-32L and music is not played.  Another World supports all of the above except the Pro Audio Spectrum and LAPC-I/CM-32L.
```

PR #275, opened 2024-09-29
Add Superfrog

tomojira:
### Superfrog

- Matches with TDC entry: `Superfrog (1994)(Team17 Software Limited) [Action]`

#### DOSC

- Copy protection patch
- Auto start


tomojira:
Revisiting this, I found that it gives a sound detection error when it starts. Changing to `run_input: y` gets past it, but trying to WAIT after that will cause the sound to stop working. So unfortunately it's stuck with a glimpse of text mode as well as the copy protection screen without being able to skip them.

PR #278, opened 2024-10-03
Add Pepper's Adventures in Time

Greyfalken:
### Source Media

- Installed from 5 3.5" floppy disks matching The Good Old Days ID: [1790](https://www.goodolddays.net/en/diskimages/?id=1790)
- Matches files installed from TDC Entry: `Pepper's Adventures in Time (1993)(Sierra On-Line, Inc.) [IMA][1440k][!]`

### Installation Options

- Graphics: VGA or IBM PS2 - 256 Colors
- Music: Sound Blaster / AdLib Card (or compatibles)
- Audio: Sound Blaster
- Joystick: Don't use a joystick
- Mouse: Use Microsoft compatible mouse
- Memory: Use your machine's extra memory

After install, delete line 8 (`directory`) in `RESOURCE.CFG`

### Official Patches

### DOSC Contents

- CPU year set to 1991
- Auto start
- Configuration file
- MT-32 and General MIDI (Roland SC-55) sound options

### Installation Variants

Install disks matching TGOD ID 1790 have been added as an installation variant.

### Digital Store Variants

### Unofficial Patch Variants


Greyfalken:
Added the file `RESOURCE.CFG` with default installation settings to the DOSZ in order to allow it to run standalone. The DOSC still contains custom versions of `RESOURCE.CFG` with specific sound mode options to overwrite the default file.

PR #289, opened 2024-10-22
Update Doom games & Chex Quest

tomojira:
- Changed to WAITMODECHANGE
- Added `cpu_max_year: 1995` to everything
- Added General MIDI option and Setup Program option to everything
- GOG and Steam variants now have the same config as the main entries to reduce complexity
- Removed DM.EXE and docs from Doom II. They weren't part of the original install, and I added them later from the Doom shareware. Sorry for changing my mind on this - I don't think they're worth the extra effort getting, people will just use IPXSETUP and SETUP for multiplayer
- Added media links
- Renamed Final Doom installation variant to id Anthology
- Finally reordered Ultimate Doom installation variant alphabetically
- Added Doom shareware as a misc. variant (it'll be fun to share!)


tomojira:
Added the original 3-episode Doom v1.9 as yet another option. Along the same lines as what we've done with some expansions, I think it might be nice to have for nostalgic value and for people who don't care about the 4th episode in Ultimate Doom.

PR #345, opened 2025-01-08
Add Secret Agent

tomojira:
### Secret Agent

- Installed v1.0 from 3.5" floppy disk matching with TGOD: [Secret Agent (ID: 829)](https://www.goodolddays.net/en/diskimages/?id=829), and also with TDC entry: `Secret Agent Trilogy [SWR] (1992)(Apogee Software, Ltd.) [IMA][720k]`
- [GOG](https://www.gog.com/en/game/secret_agent)/[Steam](https://store.steampowered.com/app/358350/Secret_Agent/) variant is added (executables are different for some reason)

#### DOSC

- Options for the 3 episodes, which use different executables
- `run_utility: true` is set for each episode
- `cpu_year: 1992` is set


Greyfalken:
I tested the Steam version and it matches with the GOG version, so you should be able to update the digital name to (GOG, Steam).

tomojira:
Great, thanks!

PR #415, opened 2025-01-22
Add Prisoner of Ice

tomojira:
### Prisoner of Ice

- Requires the [GOG release](https://www.gog.com/en/game/call_of_cthulhu_prisoner_of_ice). Redump and TDC do not have any English DOS CDs recorded for this game
- Create a DOSZ of the files in the "CD" folder - an installation variant has been added with the correct files
- [DoDAT](https://github.com/PureDOS/DoDAT) is recommended for this - RomVault/TrrntZip may not work
- Make sure the files are at the root of the DOSZ - don't include the "CD" folder itself
- Use [DOSZtoCHD](https://github.com/PureDOS/DOSZtoCHD) to build this DOSZ into a CHD file (don't use the -l or -f parameters)
- This is necessary because the game crashes when it saves if it isn't running from a CD. GOG works around this by doing `mount d "..\CD" -t cdrom`, but this can't be used in a DOSZ

#### DOSC

- `cpu_max_year: 1995` is set
- Sound Blaster 16 config
- 640x480 resolution
- Auto start


televandalist:
> Redump and TDC do not have any CDs recorded for this game

🤔 

- [Call of Cthulhu - Prisoner of Ice (Germany)](http://redump.org/disc/47109/)
- [Call of Cthulhu - Prisoner of Ice (Italy) (Rerelease)](http://redump.org/disc/77338/)
- [Prisoner of Ice (Brazil) (Es,It,Pt) (Covermount)](http://redump.org/disc/74513/)
- [Prisoner of Ice (Europe) (En,Fr,De,Nl,Sv)](http://redump.org/disc/2886/)
- [Prisoner of Ice (Taiwan)](http://redump.org/disc/95608/)

tomojira:
Ohh, I mean for the DOS version. I believe those are only for Windows. Well, I should double-check.

televandalist:
This is blowing my mind :bonk:

There's an Infogrames compilation CD with Prisoner of Ice, Solar Crusade, and Time Gate... looks like that also isn't DAT'd by Redump... or TOSEC...

tomojira:
Yep, it surprised me too. TDC claims in their notes page that there are no known good dumps - they have a bad dump which has no sound and many corrupted files.

PR #426, opened 2025-01-29
Update Warcraft I & II

tomojira:
### Warcraft I

- Added `cpu_max_year: 1994`
- Changed to WAITMODECHANGE
- General MIDI option added
- Media link added
- Comments edited

### Warcraft II

- Tides of Darkness and Beyond the Dark Portal are now split as parent and child
- [Patch v1.4](https://www.pcgamingwiki.com/wiki/Warcraft_II:_Tides_of_Darkness#Patches) was applied to Tides of Darkness. I overlooked this patch before - it's minor and mostly for multiplayer, but the readme does mention 1 bug fix
- Patch v1.5 was applied to Beyond the Dark Portal, from the same link above. It has the same minor changes
- Shared save feature is used
- CD images converted with CHDtoOGG
- Added `cpu_max_year: 1996`
- Changed to WAITMODECHANGE
- General MIDI and Sound Blaster music options added
- Media and patch links added
- Comments edited


PureDOS:
Maybe we should remove WAR2.INI from the DOSC. It would mean the player needs to disable CD audio via the in-game options but maybe that's better than having the non-default configs be weird (if selected the very first time they will have non-default settings like the intro being skipped and in-game tips being disabled, if selected afterwards they won't apply the disabling of CD audio due to WAR2.INI already existing). Do you think that's better or worse?

tomojira:
I see what you mean. It's a little weird to choose the General MIDI option, then see no effect until disabling CD music, but that seems better than the alternative. I removed the Sound Blaster option and deleted WAR2.INI from the General MIDI option.

PR #465, opened 2025-05-30
Add Immoral Cumbat

SP00RL00S:
### Immoral Cumbat

- Matches with TDC entry: `Immoral Cumbat v1.2 [SWR] (1995)(Coresoft) [Action, Adult]`

**DOSC**

- Auto start


televandalist:
I'm thinking the silly "adult" games should probably go into their own DAT since only a few people will bother with them and they'll just pollute whatever DAT they're in 😆

PureDOS:
@televandalist Probably a good idea.  Let's make a unverified/PureDOSDAT_Unverified_Adult.xml for now?

PR #496, opened 2025-08-17
Add Shadow Warrior (Steam)

Coloradohusky:
For some reason, the only difference between the GOG and Steam versions of Shadow Warrior is the first line in `GAME.DAT` - in the GOG version, it reads `FILE "GAME.GOG" BINARY`, while on Steam, it reads `FILE "GAME.IMG" BINARY`. `GAME.GOG` and `GAME.IMG` are identical apart from their names. Anyway, this PR adds support for the Steam version of Shadow Warrior.


Greyfalken:
Since everything else is the same and `GAME.DAT` only contains information on the CD image, I'd recommend moving that file to the DOSC. Then you can streamline the two digital store variants down to a single "(GOG, Steam)" version.

Coloradohusky:
Should be good now 👍

PR #498, opened 2025-08-17
Add Obitus

Coloradohusky:
### Source Media

- Installed from 3 3.5" floppy disks matching The Good Old Days ID: [1332](https://www.goodolddays.net/en/diskimages/?id=1332)

### Installation Options

- Options

Running the DOSZ is possible without the DOSC, however users will have to pass the copy protection check by reading from the manual, and can change the sound settings by running `SOUND.EXE`.

### Official Patches

### DOSC Contents

- Auto start
- Copy protection patch
- MT-32 sound option

### Installation Variants

No install media available.

### Digital Store Variants

### Unofficial Patch Variants


Coloradohusky:
Marking as draft because, going back to this, I genuinely can't tell the difference between the AdLib/Sound Blaster and MT-32 sound, so I don't know if it is active or not.

PR #517, opened 2025-10-26
Verify Ultima VII Part Two: Serpent Isle + The Silver Seed

tomojira:
My first attempt at verifying one of the games in Unverified, using the questions we came up with a while back. Please let me know if there's anything more I should do!

Yes, using [Ultima Collection (USA) (Rerelease)](http://redump.org/disc/2179/) (Redump).

Yes, but without sound at first. Run `INSTALL.EXE` to configure sound and create `SERPENT.CFG`, which is included in the DOSC.
A `DOS.YML` is included in the DOSZ to make the game playable with these settings:
```
mem_ems: false
mem_umb: false
```
Actually, only `mem_ems: false` seemed strictly necessary, but sources online say `mem_umb: false` too, so it was included just in case.
The `DOS.YML` in the DOSC redundantly included these options again, so I removed them there in this PR.

No.

Yes, `SETUP.EXE` appears to be for Windows. I removed it in this PR.

Yes, it uses CM-32L (new).
However, the `DOS.YML` was still using the old `.dosz` extension, so I updated it to `.dosx` in this PR.

No.

Yes, `cpu_year: 1993` is set, and the game speed seems okay. The water at the beginning might look a little fast, but reducing as low as 1990 doesn't change it, and GOG doesn't use any lower than that (14000 cycles).

The DOSC auto-runs `SERPENT.COM`, but `WAITMODECHANGE` is not used since it skips a chunk of the intro. The game starts at 320x200 resolution regardless.

No, though the game does create many small files in the `.pure.zip` while playing.

No, but the game does have copy protection according to PCGamingWiki - we overlooked this last time. I have added a patch to remove it to the DOSC - it affects the file `STATIC\USECODE`.

No.

Yes, the `DOS.YML` in the DOSZ already uses embedded data.

Not applicable.

Not applicable.

Not applicable.

Yes, `Ultima VII - Part Two - Serpent Isle + The Silver Seed (1993) (Origin Systems).yml` was added.

This CD-ROM version is already fully patched.

Yes, GOG matches. The game is not on Steam.

Not applicable.

Not applicable.

Yes.

Not applicable.

**Summary of changes**
- Removed redundant options from `DOS.YML` in DOSC
- Removed Windows-only file `SETUP.EXE` from DOSZ
- Updated MT-32 file extension from `.dosz` to `.dosx`
- Added copy protection removal to DOSC


PureDOS:
Seems pretty good. I think we can remove READTHIS.TXT because that's about the SETUP.EXE that was removed in the PR. Also we should remove SI.BAT and SERP.BAT (they're both pretty much the same) because in the first Ultima 7 we also removed the corresponding U7.BAT.

Did you do any further testing regarding the CPU cycles? We have set the first Ultima 7 to `cpu_year: 1990` but now this at `cpu_year: 1993`. I think I've heard that Ultima 7 is quite notorious about CPU speed (certainly emulated speed but even on real hardware). Maybe they've made improvements to the engine for part 2 to better handle faster CPUs? I think a faster CPU helps with loading times so as long as it plays as good as it can we should use that.

tomojira:
Yeah, Ultima VII Part 1 was more affected by the CPU and ran too fast. I went with `cpu_year: 1990` (14,000 cycles) there because it's what GOG used, and it appeared to be a similar speed to Part 2. Part 2 seems to have a stable speed regardless of the CPU, I tried as low as 1990 but it had no effect. So I just went with 1993 to match the year of release. I think we decided to always include the option with the year of release as a standard in cases when it isn't necessary.

In any case I removed those 3 files, and also added a couple more comments I just noticed
- Can also be extracted from CD matching with TDC#23: `Ultima Collection v1.03F (1999)(Electronic Arts, Inc., ORIGIN Systems, Inc.) [bin-cue]`
- DoDAT can build this directly from the disc image

PR #525, opened 2025-11-04
Add MDK

SP00RL00S:
### MDK

- Installed from Redump-verified CD-ROM: [MDK (Europe) (En,Fr,Es,It,Pl)](http://redump.org/disc/31515/)
- Small install
- Applied [F2 SaveKey Patch](https://3dfxarchive.com/downloads/patch-mdk.zip)

**DOSC**

- Auto start
- Sound Blaster 16 config
- WASD controls option
- `cpu_year: 1997`
- `mem_size: 32768`


televandalist:
Patch link in DAT is broken because it ends with a slash

PR #526, opened 2025-11-09
Add Archimedean Dynasty

SP00RL00S:
### Archimedean Dynasty

- Installed from Redump-verified CD-ROM: [Archimedean Dynasty (USA)](http://redump.org/disc/24453/) 
- CD image converted with CHDtoOGG (quality level `-q9`)
- Applied [1.120 and 3Dfx patches](https://www.3dfxzone.it/dir/3dfx/utilities/?objid=977) (use readme.txt from 1.120 patch)
- Added a variant with [unofficial bug fixes and tweaks](https://www.moddb.com/mods/archimedean-dynasty-augmented-mod)

**DOSC**

- Auto start
- Sound Blaster 16 config
- 3Dfx option (cockpit disabled because of emulation bug)


televandalist:
soggi.org doesn't allow hotlinking so the 1.120 and 3Dfx patches will probably need instructions to download instead of the download links

PR #550, opened 2025-11-25
Snood update

SP00RL00S:
- Updated to match with TDC entry for the registered version: `Snood v2.1 [SWR] (1998)(David Dobson) [Strategy]`
- Added `cpu_year: 1992`
- Deleted unnecessary files (PIF, ICO, RTF files + EXE file that only runs in Windows)
- Changed SNOODPRF.200 in DOSC to bps patch (the game is still shareware so I can't use the file with registration check)


PureDOS:
Do you think we can keep a copy of the old shareware version in the freeware xml file (but include your fixes)?

SP00RL00S:
> Do you think we can keep a copy of the old shareware version in the freeware xml file (but include your fixes)?

Good idea, added it to freeware.xml as `Snood (1999) (David Dobson) (Shareware)`

PR #553, opened 2025-12-08
Add blood+cryptic passage

trotsky40:
Installed from the Redump verified disc "One Unit Whole Blood (USA)"

Videos not installed (left on CD). 

Sound Blaster selected for sound with the following setting changed from default:
NumVoices = 32
NumChannels = 2
NumBits = 16
MixRate = 44000


televandalist:
- `INSTALL.LOG` should be removed entirely from the `.dosz` since it's nondeterministic
- `BLOOD.CFG` should be in the `.dosc` instead of the `.dosz`

trotsky40:
Made the changes

Issue #548, opened 2025-11-25
Dungeon Keeper Gold won't save in-game


trotsky40:
As currently listed in the DAT, Dungeon Keeper Gold gives an error and won't save in-game. The game needs a "SAVE" folder in the directory to enable saving. When installing from the source media, the game creates an empty "SAVE" folder and adding one to the output of DoDAT resolves the in-game error message. Also impacts, Dungeon Keeper- The Deeper Dungeon due to the parent/child relationship.

televandalist:
Not sure how to fix this since an empty folder will give an error in the DAT.

trotsky40:
Adding the empty folder to the DOSC for Dungeon Keeper Gold resolves for both games. I can PR it this weekend.

PR #555, opened 2026-02-20
Update to fix saving issue with Dungeon Keeper Gold and Deeper Dungeon

trotsky40:
Added an empty "SAVE" folder to fix in-game saving and added cpu_max_year: 1998 to prevent occasional speed ups


trotsky40:
Addresses https://github.com/PureDOS/DAT/issues/548

PureDOS:
Would it be better if the "SAVE" folder were added to the DAT?

trotsky40:
We didn’t think the DAT could generate an empty folder. If it can, I can rework the PR.

PureDOS:
An empty save folder can easily be added to the dat with
```xml
		<rom name="SAVE/" size="0" crc="00000000" md5="d41d8cd98f00b204e9800998ecf8427e" sha1="da39a3ee5e6b4b0d3255bfef95601890afd80709"/>
```

So having a `/` at the end of the name there means it defines a folder. This is used to create empty folders but also to have folders created with a specific date/time stamp if applicable (by also adding the `date` field). There's quite a few like these in the DAT file we have.

trotsky40:
Copy. I will rework it.

trotsky40:
Moved the save folder to the dat

PureDOS:
Does the GOG variant in PureDOSDAT_Variants_Digital.xml also need such an entry?

trotsky40:
Took me a bit of time to test. The GOG Variant also needed the folder. Now added.

PR #560, opened 2026-05-16
Dune II update

trotsky40:
Updated Dune II with an unofficial bug patch (available through TDC). Retains MT-32 audio fixes and multiple languages. The game still has copy protections (available cracks only work on US version without the audio fixes and multiple languages). Manual for copy protection questions can be found here: https://archive.org/details/dune-ii-manual/page/n5/mode/2up


PureDOS:
This is this patch https://forum.dune2k.com/topic/20168-dune-2-v107-fix/ ?

This does not seem to be in line with the project description which states:

> Modern patches or updates made long after the original DOS release should not be included.

Do you think this patch is a requirement for Dune 2 to be playable?

PR #614, opened 2026-05-31
Add Populous II: Trials of the Olympian Gods

Coloradohusky:
### Source Media

- Installed from Redump-verified CD-ROM: [51514](http://redump.org/disc/51514/)
- Installed from Redump-verified CD-ROM: [80432](http://redump.org/disc/80432/)
- Installed from 4 3.5" floppy disks matching The Good Old Days ID: [1832](https://www.goodolddays.net/en/diskimages/?id=1823)
- Matches TDC Entry: [Populous II- Trials of the Olympian Gods (1993)(Electronic Arts, Inc.) [Strategy]](https://puredos.github.io/tdc-browser/?/23/Populous%20II-%20Trials%20of%20the%20Olympian%20Gods%20(1993)(Electronic%20Arts,%20Inc.)%20%5BStrategy%5D)

Unable to build with DoDAT directly from Redump or TGOD sources, so install script must be run first.

### Installation Options

Configuration is contained in the DOSC instead of the DOSZ, so installation options do not matter. To play the game without the DOSC, users can run `INSTALL.EXE` (from the install media) to produce `GO.BAT`, or can simply run `POP2.EXE` with `/h` for High-Res, `/l` for Low-Res, and `/ss` for Sound Blaster (such as `pop2 /h /ss`).

### Official Patches

### DOSC Contents

- Auto start
- CPU year set to 1993
- Two video options, for High-Res 16-Color and Low-Res 256-Color

### Digital Store Variants

Unable to test GOG/Steam versions, as I do not own them.

### Unofficial Patch Variants


televandalist:
Instances of `&` should be replaced with `&amp;`

Edit: Out of the Park, too.

PR #632, opened 2026-06-08
Add Judge Dredd Pinball

Coloradohusky:
### Source Media

- Installed from Redump-verified CD-ROM: [83194](http://redump.org/disc/83194/)

### Installation Options

Configuration is contained in the DOSC instead of the DOSZ, so installation options do not matter. To play the game without the DOSC, users can run `INSTALL.BAT` (from the install media) to produce `GO.BAT`.

### Official Patches

### DOSC Contents

- Auto start
- SoundBlaster 16 config file
- Copy protection patch
- Languages ENGLISH (default), FRENCH, GERMAN, ITALIAN, and SPANISH

### Digital Store Variants

### Unofficial Patch Variants


Coloradohusky:
discovered it doesn't work with the Europe Redump, investigating

Coloradohusky:
okay should be good now, there was a one-byte difference in DREDD\MB800\GFX.WAD between the Europe/Australia regions, compared to the German region. If using the German install, users can change byte 0x37 to 0x33 at offset 0x153DFE to build the DOSZ

televandalist:
DAT still says it uses the European version instead of the German version

Also, it took a bit to get the cue file to match since the region was omitted.

Coloradohusky:
Ah yeah sorry, I forgot to add the CD audio the first time around - so I switched to just the Europe Redump for the audio. Game is playable without the CD audio though

I'll fix the XML tonight

Coloradohusky:
does this CUE match what it should now?

PR #637, opened 2026-06-13
Verify Ultima VIII: Pagan

tomojira:
This game had 6 entries added - English, French, German, GOG English, GOG French, and GOG German. Plus the Installation variant

No, I added more explanation. Running a second installer is needed after the first to install the Speech Pack.

```
Installed using Redump-verified CD-ROM: Ultima VIII - Pagan (USA) (En,Fr,De) (Rerelease)
Run "install english" at command line in D:
Install options:
	Source Drive: D:\ENGLISH (Default)
	Destination Drive: C:\ULTIMA8 (Default)
	Sound Card: Sound Blaster Pro, 16/16ASP - Port 0x0220, IRQ 7, DMA 1
	Music Card: Sound Blaster Pro - Port 0x0220
Then run D:\ENGSPCH\INSTALL.EXE to install the Speech Pack
DoDAT cannot build this directly from the disc image
```

The GOG comments mention a problem with `AILXMI.DLL`, but I didn't encounter it. `AILXMI.DLL` appears to be identical to `SOUND/A16SBFM.DLL` (or `SOUND/A16MT32.DLL` if you use GM).

This CD-ROM release is already fully updated to v2.12, and also includes the Speech Pack.

Yes, there were 3 GOG variants (English, French, and German) added. The English GOG variant actually matched with the main entry after removing only 1 file (`README.TXT`), so I did that and eliminated it. The same couldn't be done with the French/German GOG variants and the main French/German variants because they have a different `INSTALL.EXE`, needed to change sound config.

I didn't try to combine the main French/German variants with the English one (using DOSC options for language instead) because they each have a different `U8SAVE.000` file, and it contains quite a lot (410 KB).

No.

No, the files on the disc are in containers that are unpacked by the installer.

Not applicable.

No.

The file timestamps are unknown. The files on the disc are in containers.

Yes, but you have to run `INSTALL.EXE` first to generate `U8.INI` sound config.

Not applicable.

Not applicable.

No, though several files are created when playing.

No.

Not applicable.

`cpu_year: 1994` was set, but I changed it to `cpu_max_year: 1994` because the cycles were dynamic. The speed seems fine.

Yes.

There is no MT-32 support.

Yes, I added it. An XOR patch is used to copy `SOUND/A16MT32.DLL` to `AILXMI.DLL` (the filename says MT32 but it is GM).

Not applicable.

Not applicable.

Yes, I added it.

Not applicable.

Not applicable.

**Summary of changes**
- Edited comments and links
- Removed `README.TXT` from the main entry, allowing GOG to match. Therefore removed GOG English variant
- Changed from `cpu_year: 1994` to `cpu_max_year: 1994`
- Added General MIDI option
- Added `input_directmouse: true`

Unmentioned above:
- Changed the Sound Blaster config to IRQ 7 from 5
- Removed a line in the main German variant DOS.YML that seemed to be an error: `run_mount: CD\U8.CHD`


PureDOS:
According to the [file list](https://www.gogdb.org/manifest/cdcb1e13cf45fe2eee43bac5a4da3335) on the [game page on GOGDB](https://www.gogdb.org/product/1207659015/build/52190466263847106) the build on GOG includes README.TXT. Can we add it back in for all variants?

Also, besides INSTALL.EXE, should we make the file listing of the GOG language variants match the other variants? It seems just a sloppy release on GOG where English files have been mixed with the other languages (reuse of the English INSTALL.EXE instead of the localized ones and having the E*.* files exist even though there should be just G or F files). So we could make "(GOG, French)" an identical copy of "(French)" but with INSTALL.EXE replaced with the 30276 byte sized entry from the base English variant. Or we just keep the English INSTALL.EXE in all variants and avoid the need to have GOG variants at all. I think that would be acceptable.

tomojira:
1. Yes, sorry about that, I overlooked README.TXT because it was in GOG's base folder instead of the ENGLISH folder.
2. I put the English INSTALL.EXE in the French and German variants and eliminated the GOG variants.

PR #640, opened 2026-07-21
Verify King's Quest VI: Heir Today, Gone Tomorrow

tomojira:
## Verifying King's Quest VI: Heir Today, Gone Tomorrow (1992) (Sierra On-Line)

This game is already in the verified dat, but double-checking it since it hasn't been done before.

### Checklist

No. I added installation instructions to reproduce the `RESOURCE.CFG` file, which is required to play the game; it can't be created later by `INSTALL6.EXE` because it will be missing a "minHunk" value which is apparently only added during installation, and the game will give an error if it isn't there.

This `RESOURCE.CFG` is slightly different than the current one, even though options are the same, so the copy in the DOSC and embedded data were both updated.

No patches are mentioned on PCGamingWiki. This is likely the latest version, from the King's Quest Collection CDs.

Yes, the GOG variant works. However, I had to edit its `RESOURCE.CFG` to change its audio driver to Sound Blaster. Added a copy of it to the DOSC as well as embedded data, and instructions in the dat comment.

Yes, and it already does so with `RESOURCE.CFG`.

Yes, but the Gabriel Knight Demo is required as well to provide its `VGA320.DRV` file, which fixes stuttering in the intro video. This is already explained in the dat comment, and a link is provided.
https://archive.org/details/GabrielKnightSinsoftheFathers_1020

Not applicable.

No.

Yes.

Yes, with `RESOURCE.CFG` included.

No. The CD includes an audio track which plays during the game according to a quick search.

Yes, and it already does so.

No.

No.

The game has copy protection according to PCGamingWiki, but unclear if that applies to this CD version. TDC's notes mention cracking a file that doesn't seem to exist here.

The speed seems fine, but there is no CPU year specified. I added `cpu_year: 1992`.

Yes.

Yes, and it's correctly using Old. The GOG variant was missing this so I added it there.

Yes, but it wasn't in the DOSC yet. I added it.

Not applicable.

Not applicable.

Yes, I added it.

Not applicable.

Not applicable.

Yes.

Yes.

There is a "Small" installation size, but it installs nothing at all besides `RESOURCE.CFG`. The "Large" option adds `INSTALL6.EXE` for changing settings.
EDIT: I realized this explanation makes no sense, since `INSTALL6.EXE` can also be used from the CD. However, after testing the Small version, it is significantly slower, taking several more seconds for the game to start after hitting Play, and also taking more time when walking to a different area. So Large seems better here.

**Summary of changes**
- Added installation instructions to dat comment. Edited comments in general
- Changed to a slightly different `RESOURCE.CFG` file which is closer to what you get after installation, so minimal editing is needed. Updated embedded data and the copy in DOSC
- Added `cpu_year: 1992`
- Fixed GOG variant to use Sound Blaster driver, and added MT-32 option
- Added General MIDI option
- Enabled direct mouse


PureDOS:
Now that we have [ExtractDATEmbeds](../../ExtractDATEmbeds) for users that don't use DoDAT, I think we can remove the base RESOURCE.CFG from the DOSC. Other than that this is a great update, thanks!

PR #669, opened 2026-08-13
Verify Ultima 0: Akalabeth

tomojira:
## Verifying Ultima 0 - Akalabeth (1998) (Origin Systems)

The main dat version of this game requires CWSDPMI r7 to fix a crash after choosing a lucky number - the comment mentions this, but the file list was accidentally still using the old version, so I updated it. It also needs a BAT file to run CWSDPMI with the right option - this was already in the DOSC, but I added it to the DOSZ as well so it's more playable on its own. Some info on this:
https://www.dosbox.com/comp_list.php?showID=717&letter=A

This does not apply to the GOG variant - it doesn't use CWSDPMI at all.

### Checklist

Config files are not needed in the DOSZ - `SETUP.EXE` can create it at any time. The GOG variant had `SOUND.CFG` left in, but I removed it.

There are no patches.

Yes.

Not applicable.

Yes.

Yes, the GOG variant has already been added.

Yes, I added embedded data along with `AK.BAT`.

Yes.

Not applicable.

Not applicable.

No, I added them to the main dat version.

Yes.

Not applicable.

Not applicable.

No.

No.

Not applicable.

`cpu_year: 1983` was already in the main dat version. It was missing from the GOG variant, so I added it there.

Yes.

Not applicable.

No, the option is there but it doesn't seem to work.

Not applicable.

Not applicable.

Not applicable.

Not applicable.

Not applicable.

**Summary of changes**
- Updated to CWSDPMI r7 in main variant
- Added `AK.BAT` with embedded data to main variant DOSZ
- Added file timestamps to main variant
- Removed `SOUND.CFG` from GOG variant DOSZ
- Added `cpu_year: 1983` to GOG variant
- Changed year of release from 1980 to 1998 - yes, this game was never officially ported to DOS until then

EDIT: Later update
- Removed `AK.BAT` from DOSC due to redundancy with embedded data


PureDOS:
This CWSDPMI business seems a bit awkward but I'm not sure if that's somehow more faithful to what was originally released than what is on GOG. But seeing that the GOG version is also the same "Akalabeth Rev 2 v.0910" with all data files being the same and just the EXE files being smaller, and it is being given away for free, I see no reason why we shouldn't just have that one version. What do you think?

tomojira:
I was reluctant to do that since I don't know where GOG's version came from, or if they modified it themselves. It also requires creating a GOG account even though it's free. But it's also strange that the original version doesn't work with the old CWSDPMI it came with, so I can do it.

PureDOS:
Ah, I figured it out. Well, it was already written in the second comment on [the dosbox.com page you linked](https://www.dosbox.com/comp_list.php?showID=717&letter=A). So this is how you can get the simpler GOG EXEs from the CWSDPMI depending EXEs from the Ultima Collection CD:
1. Go to https://tippach.hier-im-netz.de/wdosx/
2. Get [wdosx097.zip](https://tippach.hier-im-netz.de/wdosx/wdosx097.zip)
3. Create a folder, extract STUBIT.EXE from WDOSX into it
4. Also put in AK.EXE (572510 bytes) and SETUP.EXE (328192 bytes) next to it
5. Launch/mount that folder in DOSBox
6. Run `STUBIT AK.EXE` and `STUBIT SETUP.EXE` (will write AK.EXE with 238962 bytes and SETUP.EXE with 190043 bytes)

So the EXEs on GOG are the same EXEs with a different DOS extender that can embed itself in an EXE and then also compresses itself (thus the smaller filesize).

tomojira:
Ohh, good to know! Should we include these instructions in the dat comment? Or just leave it at GOG?

PureDOS:
Probably a good idea. Doesn't need to be that verbose. I think we can just say "It is also possible to build the game from the Ultima Collection CD-ROM by running AK.EXE and SETUP.EXE through STUBIT from WDOSX 0.97" and add a
`<link type="patch">https://tippach.hier-im-netz.de/wdosx/</link>` or something?

tomojira:
Okay, great! I also re-added the timestamps for most of the files (should the 2 patched files have them too?)

PureDOS:
Yeah, I'd probably put just the dates we got from the CD-ROM onto the WDOSX patched EXEs, it looks nicer :-)

tomojira:
Okay, done!

PR #670, opened 2026-08-13
Verify Ultima I: The First Age of Darkness

tomojira:
### Checklist

Not applicable - there are no config files.

Not applicable.

Yes.

Not applicable.

Yes.

Yes, the GOG version matches with the main variant.

No.

Yes.

Not applicable.

No.

Yes.

Yes.

Not applicable.

Not applicable.

No.

No.

Not applicable.

Yes, `cpu_year: 1981` slows down the game.

No, I added (WAITMODECHANGE) to skip to the 320x200 change. Oddly, this wasn't needed the last time this game was checked.

Not applicable.

Not applicable.

Not applicable.

No.

Not applicable.

Not applicable.

Not applicable.

**Summary of changes**
- Added (WAITMODECHANGE)


PureDOS:
How do you feel about setting it to `cpu_year: 1982`? To me it feels a bit nicer especially walking through the world and dungeons. Sound effects still sound ok to me. What do you think?

tomojira:
Sure, seems good.

PR #682, opened 2026-08-14
Verify King's Quest V: Absence Makes the Heart Go Yonder!

tomojira:
## Verifying King's Quest V - Absence Makes the Heart Go Yonder! (1990) (Sierra On-Line)

This game is already in the verified section, but just giving it another look with the checklist since it was added early on.

### Checklist

Main variant: I added instructions on how to get `RESOURCE.CFG`, which is required in the DOSZ.
I noticed the current version had some different parameters than the one you get from installation, so I changed it. May be unnecessary, but just in case.

GOG variant: There was no copy of `RESOURCE.CFG` in the DOSC, so I added it there.

No official patches according to PCGamingWiki.

Yes.

Not applicable.

Yes.

Yes, there is already a GOG variant.

Main variant: Yes, it's already used for `RESOURCE.CFG`.

GOG variant: It was missing, so I added it.

Yes.

Not applicable.

No.

Yes.

Yes.

Not applicable.

Not applicable.

No.

No.

There is no copy protection for the CD version according to PCGamingWiki.

No CPU year was specified. I added `cpu_year: 1990` to both variants since there didn't seem to be speed issues at the beginning.

Yes. Added a longer WAIT to the MT-32 option since it takes significantly longer to load.

Main variant: Yes, it uses "old".

GOG variant: It was missing, so I copied it over.

No.

Not applicable.

Not applicable.

Yes, I added it to both variants.

Not applicable.

Not applicable.

**Summary of changes**
- Main variant: Changed to installed version of `RESOURCE.CFG` in both DOSZ and DOSC, just in case, with only paths edited
- GOG variant: Added `RESOURCE.CFG` to DOSC
- GOG variant: Added embedded data to `RESOURCE.CFG` for DOSZ
- Both variants: Added `cpu_year: 1990`
- GOG variant: Added MT-32 option
- Both variants: Increased WAIT for MT-32 option
- Both variants: Enabled direct mouse
- Edited comments and links

Other changes not mentioned in the checklist:
- Removed 3 unnecessary files from GOG DOSZ

EDIT: Later update
- Removed `RESOURCE.CFG` from both DOSCs due to redundancy with embedded data


PureDOS:
The RESOURCE.CFG embedded in ‎PureDOSDAT_Variants_Digital.xml for "King's Quest V - Absence Makes the Heart Go Yonder! (1990) (Sierra On-Line) (GOG).dosz" as well as under `[MT-32 Sound]` in "King's Quest V - Absence Makes the Heart Go Yonder! (1990) (Sierra On-Line) (GOG).dosc" contain ` directory = \SIERRA\kq5` which I think should be ` directory = .\`, right? It also is missing `audio = .\` even though AUDIO001.002, which I think is the speech data, exists.

tomojira:
This is because it's using the default RESOURCE.CFG file provided by GOG. The directory parameter actually doesn't seem to matter with many Sierra games, and could even be deleted (but I didn't remember that until after editing them to .\ instead). I'm not sure if the audio parameter matters for this game, since speech does seem to be playing in-game regardless. In any case, if GOG users are assumed to be using DoDAT, it doesn't really matter if it's the default file, or if it's changed to be consistent with the main one, it will be generated either way. The default could be more convenient for RomVault. But I'm not sure which way is best.

PR #696, opened 2026-08-17
Verify Police Quest 3: The Kindred

tomojira:
## Verifying Police Quest 3 - The Kindred (1991) (Sierra On-Line)

EDIT: I previously verified the version from the Police Quest Collection CD, but I have regretted it since it doesn't include the standard `INSTALL.EXE` for changing config, and the replacement `INSTALL.BAT` works oddly and imperfectly.

Changed to a better version, from floppy disks matching with TDC#23 entry: [Police Quest 3- The Kindred (1991)(Sierra On-Line, Inc.) [IMA][1440k][!]](https://puredos.github.io/dat-browser/?/Police+Quest+3-+The+Kindred+(1991)(Sierra+On-Line,+Inc.)+%5BIMA%5D%5B1440k%5D%5B!%5D)
Also matches with TDC#23 entry: [Police Quest 3- The Kindred (1991)(Sierra On-Line, Inc.) [Adventure]](https://puredos.github.io/dat-browser/?/Police+Quest+3-+The+Kindred+(1991)(Sierra+On-Line,+Inc.)+%5BAdventure%5D)

### Checklist

I added instructions for reproducing `RESOURCE.CFG`. It's needed in the DOSZ because the necessary `minHunk` parameter isn't created by `INSTALL.EXE` later.

A couple patches are mentioned by PCGamingWiki, but they don't sound necessary.

Yes.

Not applicable.

Yes.

Yes, a GOG variant was already added. Unfortunately it no longer matches since a GOG update, so I changed it.

Both variants: Added embedded data to `RESOURCE.CFG`.

Yes.

Not applicable.

No.

Yes.

Yes.

Not applicable.

Not applicable.

No.

No.

Main variant: Unknown if this version has copy protection. Nothing appeared at the beginning of the game.

GOG variant: Doesn't have copy protection according to PCGamingWiki.

Both variants: No CPU setting was specified. I added `cpu_year: 1988` to slow it down in hopes of preventing speed bugs.
https://sierrahelp.com/Games/PoliceQuest/PQ3Help.html#Speed

Both variants: No, I added WAITMODECHANGE. MT-32 has a longer WAIT as well.

Main variant: Yes, it already uses "old". The extension was still .dosz, so I updated it to .dosx.

GOG variant: It was missing, so I added it here.

No.

Not applicable.

Not applicable.

Both variants: Yes, I enabled it.

Not applicable.

Not applicable.

**Summary of changes**
- Main variant: Installed different version from floppy disks matching with TDC#23 entry: [Police Quest 3- The Kindred (1991)(Sierra On-Line, Inc.) [IMA][1440k][!]](https://puredos.github.io/dat-browser/?/Police+Quest+3-+The+Kindred+(1991)(Sierra+On-Line,+Inc.)+%5BIMA%5D%5B1440k%5D%5B!%5D)
- GOG variant: Updated version
- Both variants: Added embedded data to `RESOURCE.CFG`
- Both variants: Added `cpu_year: 1988`
- Both variants: Added WAITMODECHANGE, as well as a longer WAIT for MT-32
- Main variant: Updated MT-32 extension to .dosx
- GOG variant: Added MT-32 option
- Both variants: Enabled direct mouse
- Edited comments and links

Unmentioned in the checklist:
- Added new installation variant

EDIT: Later update
- Removed `RESOURCE.CFG` from both DOSCs due to redundancy with embedded data


PureDOS:
This one took me a while to figure out...

The [Steam version](https://steamdb.info/depot/494741/?show_hashes) matches the build that [used to be available on GOG before](https://www.gogdb.org/manifest/413f72b3-e31b-462f-927c-b365abb34355) (which has [since been updated](https://www.gogdb.org/manifest/013003b253437ac59941e13e5a5b344e)). That release basically matches the 1.44 MB floppy version except it contains a few more patch files:
- 202.SCR, 202.TEX from the "Fix Pentagrama (PQ3PENT)" patch listed on [PCGW](https://www.pcgamingwiki.com/wiki/Police_Quest_III:_The_Kindred) ("patch is to fix the pentagram problem found on PQ3 if using a keyboard")
- 99.SCR, 994.SCR from the "fix (English version)" patch listed on an [old GEOCITIES page](https://www.oocities.org/belzorash/) ("Fixes: Car driving sequences, Some animations")

I'm not sure if these patches were originally created by players but I think they made it onto a latter official release like the CD-ROM version. By using a speed limited emulator these patches are potentially unnecessary.

Because the current version on GOG matches the main variant, it doesn't need to be separate (it even includes INSTALL.EXE this time). But I added a Steam variant now but I did exclude the additional patch files to make the two variants identical in terms of game content. Besides missing INSTALL on the Steam variant, two sound drivers (SNDBLAST.DRV and MTBLAST.DRV) seem to slightly differ. These are exactly the sound drivers we use so I think there is a chance that this would affect how data is stored in RAM between the variants. Alternative would be to base the main variant on the CD-ROM release but then GOG needs to be a variant and it ends up the same.

